PR #4, opened 2024-07-02
Use freelens-k8s-proxy

robertobandini:
I renamed where necessary and tested the build.


robertobandini:
For now we can still move forward, I tested the local build on Windows 10, we will fix the tests/workflow later.

Issue #10, opened 2024-08-08
Pipelines & Release workflow


robertobandini:
Fix the actions needed to re-establish a development and release workflow

dex4er:
Hi. I'm still working on it. Unit tests now are fine, integration tests still break on Ubuntu: https://github.com/dex4er/freelens/actions

I would like to introduce nightly builds and I think they should be built on separate repo to not confuse normal users who will be looking for normal releases.

Would you mind to create repo like freelens-nightly or something similar?

dex4er:
Integration test breaks on Ubuntu on rpmbuild command. This is not native environment for rpm but somehow lens was provided as rpm package too, then I'll rather investigate it than disable rpm as a electron build package.

robertobandini:
> freelens-nightly

https://github.com/freelensapp/freelens-nightly-builds

robertobandini:
Hi @dex4er you have now write permissions the needed repo, so I think you are able to go on working on the release process.
Can you help us to complete it about this issue and also about this https://github.com/freelensapp/freelens-node-pod-menu/issues/32 even without signign?
Meanwhile I'll check out too that topic:
https://github.com/freelensapp/freelens/issues/13
It's also important to find other collaborators to work on other imporetant topics like fixing the security issues.

dex4er:
This is the last step: to adapt the current nightly and make a final release.

PR #11, opened 2024-08-10
CI: New workflows for tests

dex4er:
Partial work for #10 

It removes all workflows and recreates for tests only: unit tests, integration tests, and linter.
The others were related to the releases (it'll be recreated again in separate PR) and handling issues/labels, however this project has only the default labels so workflows fail.

I would like to (re)introduce https://github.com/release-drafter/release-drafter in a separate PR. Probably we don't need to automate releases now when the rebrading is not yet finished.

unit and integration tests are run on 3 all platforms (maybe it is not important for unit tests?)

Now, the default for `build:app` is to build only for the current OS, then it is possible to add the explicit target. It speeds up integration test as the target might be `dir` for current arch only by adding explicit CLI arguments for `electron-builder`.

linter test is run only on Ubuntu.

Also, PR removes all editor-related directories: both .idea and .vscode. I.e., I use VScode and don't think forcing anybody with my local configuration would be a good idea.

As a bonus: the main screenshot is replaced and presents the current version of Freelens.

Please squash it.


dex4er:
Ready to merge

robertobandini:
Nice job, but I think it's also important to reintroduce Dependabot&Renovatebot so I created this https://github.com/freelensapp/freelens/issues/15
Maybe we can work to this after pipelines/rebranding, security alerts are importants, what do you think?

PR #18, opened 2024-09-23
Add Freelens author were it is missing

robertobandini:


robertobandini:
I had also to correct the package.json inside freelens dir and I simply did it through a commit directly on the main branch.

robertobandini:
@dex4er fyi

PR #27, opened 2024-10-03
CI: workflow to release the packages

dex4er:
It is triggered for git tags. It creates packages for each target. It is almost the same as for nightly builds except here after publishing GitHub release there is also published to npmjs.com.


dex4er:
The logic here is to pubish NPM packages if binary packages are built so we should be more or less in-sync between GitHub and NPM releases.

Unfortunately I still fight with lerna and some problems that prevents `lerna publish` like ie.:

```
lerna notice cli v6.6.2
lerna info versioning independent
lerna ERR! EUNCOMMIT Working tree has uncommitted changes, please commit or remove the following changes before continuing:
lerna ERR! EUNCOMMIT M  packages/kubectl-versions/build/versions.json
```
or more problematic:
```
lerna ERR! E429 Could not publish, as user undefined: rate limited exceeded
```

dex4er:
There is ~50 packages to publish and it looks like npmjs.org limits us after 25 packages.

In worst case we wouldn't be able to publish NPM packages from the CI pipeline or eventually we could switch to ie. GitHub registry only.

It would be possible to have the NPM packages in Github as they can be per @namespace:

```
@NAMESPACE:registry=https://npm.pkg.github.com
```

robertobandini:
> There is ~50 packages to publish and it looks like npmjs.org limits us after 25 packages.
> 
> In worst case we wouldn't be able to publish NPM packages from the CI pipeline or eventually we could switch to ie. GitHub registry only.
> 
> It would be possible to have the NPM packages in Github as they can be per @namespace:
> 
> ```
> @NAMESPACE:registry=https://npm.pkg.github.com
> ```

Isn't unlimed for free public packages?
I saw that now there are 28 packages under the freelensapp organization on npm.

dex4er:
All 51 are now published however it looks like it is not possible to do it in one shot, ie. from GitHub Actions

robertobandini:
Hi @dex4er I did some research, but I can't find any documented rate limit or difference between npm free/pro/team account about this problem.
Until now I've only found some page similar to this that let understand me that maybe there is some bug or temporary problem:
https://stackoverflow.com/questions/75973455/npm-publish-429-too-many-requests
I think that the original workflow published all the packages without problems:
https://github.com/lensapp/lens/blob/master/.github/workflows/publish-release-npm.yml
Could you please check more in detail? Or maybe just retry.
If you need, just let me know, and I'll try myself to to try and verify more.

dex4er:
I had exactly what I wrote in previous command:

```
lerna ERR! E429 Could not publish, as user undefined: rate limited exceeded
```

after I published ~25 packages. Then it occurred for at least next 24 hours.

Maybe free account has some daily limits, maybe something else. The publishing is done in separate step then we can just retry this action separately from the building or publishing to GitHub.

robertobandini (approved):
Yes, we should verify, anyway also the original release workflow had even a separate action to publish npm packages.

PR #33, opened 2024-10-13
Update bundled binaries

dex4er:
* kubectl 1.31.1
* helm 3.16.2
* k8s-proxy 1.1.0

The default URL for kubectl is changed to https://dl.k8s.io/release/ (current official) as the old one does not have newer versions. It should fix the problem that bundled version of kubectl is used:

```
freelens: 2024-10-15T17:50:50.769Z pw:browser [pid=12716][out] info:    ┏ Downloading kubectl 1.30.5 from https://storage.googleapis.com/kubernetes-release/release/v1.30.5/bin/linux/amd64/kubectl to /tmp/lens-integration-testing/991f4e68-049c-4802-92f1-3812f7259c26/Freelens/binaries/kubectl/1.30.5/kubectl +42ms
freelens: 2024-10-15T17:50:50.769Z pw:browser [pid=12716][out] info:    ┗ [1] { timestamp: '15/10/2024 17:50:50' }
freelens: 2024-10-15T17:50:50.837Z pw:browser [pid=12716][out] error:   ┏ [KUBECTL]: failed to download kubectl Response code 404 (Not Found) +68ms
freelens: 2024-10-15T17:50:50.837Z pw:browser [pid=12716][out] error:   ┃ [ 1] HTTPError: Response code 404 (Not Found)
freelens: 2024-10-15T17:50:50.838Z pw:browser [pid=12716][out] error:   ┏ Failed to ensure kubectl, fallback to the bundled version +1ms
```

Additional patch:

* Binaries downloader shows full URL now.

Closes #32

PR #35, opened 2024-10-15
Extra setttings for MacOS PKG

dex4er:
It'll be used when `npm run build:app -- -- -- pkg` runs.


dex4er:
So it breaks because of #33 so I think I'll prepare separate PR for clarity.

dex4er:
https://github.com/freelensapp/freelens-nightly-builds/actions/runs/11352359439/job/31574945358 shows if builds correctly

PR #45, opened 2024-10-20
CI: Skip pushing/PR if no changes

dex4er:
Correct workflow for kubectl updates


robertobandini:
Great work @dex4er, anyway I'm thinking would be useful to set also the description in every github action, so everyone at first sight can know what it does beyond the its title... what do you think?

PR #61, opened 2025-01-08
The DEB package for arm64 misses binaries for kubectl, helm and k8s-proxy

dex4er:
x64 has them, arm64 not.


dex4er:
Ok so this is regression as we now miss `DOWNLOAD_ALL_ARCHITECTURES=true` for macOS and Linux.

PR #104, opened 2025-01-27
Added advertisement of @freelensapp/freelens-node-pod-menu

dex4er:
This is a recommendation of `@freelensapp/freelens-node-pod-menu` so it should be easier to install it by new users.


robertobandini:
Hi @dex4er I agree could be useful.
We can even consider to install it by default in some way while waiting to include it again directly in Freelens.

PR #109, opened 2025-01-28
Fix for incorrect launchable tag in metainfo file

dex4er:
FIxes incorrect `<launchable>` tag as it was discovered in https://buildbot.flathub.org/#/builders/6/builds/177149:

```json
{
    "errors": [
        "metainfo-launchable-tag-wrong-value"
    ],
    "info": [
        "metainfo-launchable-tag-wrong-value: The value of launchable tag in Metainfo is wrong: app.freelens.Freelens"
    ],
    "message": "Please consult the documentation at https://docs.flathub.org/docs/for-app-authors/linter"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,17 @@ flatpak install flathub app.freelens.Freelens
 flatpak run app.freelens.Freelens
 ```
 
+The application is sandboxed. It includes bundled `kubectl` and `helm`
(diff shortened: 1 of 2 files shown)


dex4er:
I'm not sure if it is so important that we should release a new package soon.

PR #110, opened 2025-01-30
More tweaks for metainfo

dex4er:
More screenshots, better description and brand colors.

image
image

Recommendation from Flathub:

image
image


robertobandini:
Great, thanks!

dex4er:
Hmm, I wait for another verification on the Flathub. We'll see if we miss something else.

PR #128, opened 2025-02-03
Regenerated icons for all OSes

dex4er:
Another attempt to make the correct icons everywhere.


dex4er:
Hmm it looks like electron-builder detects sizes incorrectly:

https://github.com/freelensapp/freelens/actions/runs/13118216641/job/36597658115?pr=128#step:16:39

PR #134, opened 2025-02-09
node-pod-menu extension is a part of the main application

### LINKED ISSUE #129: Install the freelens-node-pod-menu extension in Freelens by default
**What would you like to be added**:
The "freelens-node-pod-menu" extension should be installed by default in the Freelens application.

**Why is this needed**:
Its features, namely log view and exec for containers, are considered essential.
This is why it was already planned to reintegrate them directly into the application without having to install an extension.
It should therefore also be evaluated whether it is better to proceed directly to reintegrate them into the application or to first install the extension automatically by default and then subsequently reintegrate it into the core.

mariomamo:
- Fixed containers status color on "Log" menu
- Added unit tests

This pull request resolve #129


robertobandini:
Great work @mariomamo 🎉🎉🎉

PR #136, opened 2025-02-09
Provide native icon.icns for macOS

dex4er:
The icon saga continues. This time is an explicit convertion of the SVG to the native icns format.

Now the sizes look good:

image

image


robertobandini (approved):
Thanks! It looks good! If needed in future I can involve also friends that are expert about UI/UX/graphics.

robertobandini (approved):

PR #142, opened 2025-02-12
Use annotation rather than label for marking manual edits

### LINKED ISSUE #141: Rebranding of resource label k8slens-edit-resource-version
**Describe the bug**
After editing any resource the label k8slens-edit-resource-version is added. It should be rebranded as well.

Maybe we should also use annotation rather than label to not break other IaC tools (Terraform, Flux).

**To Reproduce**
1. Go to any resource
2. Edit/Save
3. New label appears

dex4er:
Fixes #141

**Description of changes:**

- Replaces label with annotation
- Renamed to keep convention `foo.domain.name/something`


dex4er:
`freelens.app/resource-version` annotation looks as the best option

PR #146, opened 2025-02-13
Defining the mountpoints per cluster for the filesystem metrics

### LINKED ISSUE #145: The graph about node disk usage shows only root disk
**Describe the bug**
The Freelens shows only metrics for the root disk. It makes this feature useless if the node uses immutable filesystem like Bottlerocket OS.

In this case the usage is always the same even if the state is stored on another disk:

**To Reproduce**
1. Go to Nodes
2. Click on node that runs Bottlerocket OS
3. Click on Disk graph

**Expected behavior**
Real disk usage should be presented

**Screenshots**
Image

**Environment (please complete the following information):**
- Freelens Version: 0.1.4

dex4er:
Fixes #145

### DIFF
--- a/packages/core/src/common/cluster-types.ts
+++ b/packages/core/src/common/cluster-types.ts
@@ -178,7 +178,7 @@ export enum ClusterMetricsResourceType {
 /**
  * The default filesystem mountpoints for metrics
  */
-export const initialFilesystemMountpoints = "/";
+export const initialFilesystemMountpoints = "/|/local";
 
 /**
  * The default node shell image
--- a/packages/core/src/common/cluster-types.ts
+++ b/packages/core/src/common/cluster-types.ts
@@ -125,6 +125,7 @@ export interface ClusterPrometheusPreferences {
   prometheusProvider?: {
     type: string;
   };
+  filesystemMountpoints?: string;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


dex4er:
I'm not sure if we should leave `/` as default or set it to `.*`

image

@robertobandini Do you have any AKS for verification maybe? I tested it on EKS with Ubuntu and Bottlerocket OS.

Before:

Ubuntu

image

Bottlerocket

image

After:

Ubuntu

image

Bottlerocket

image

dex4er:
Ok, so the finding is that '/` is too less and `.*` is too much. node-exporter duplicates some filesystems if they are mounted many times then numbers are summed incorrectly. In my case it should be ~1TB but the same filesystem is bind-mounted 4 times.

The safe default would be `/|/local|/mnt|/mnt/.*|/var` which should cover the most important systems (Bottlerocket, COS, Talos. Ubuntu)

robertobandini:
> Uh, `/var` and `/mnt` are doubled on Bottlerocket too then we won't have universal regexp for every OS

Ok, so If I understood, we need a different value depending on the OS, isn't it?
Maybe we can autodetect it?
At first glance in summary on AKS the old default value works on Linux nodes, but if needed I can do an in-depth test.

Note: to see the disk metrics I had to put Helm instead of Auto detect; it seems to me that something needs to be revised regarding auto detect, what do you think?

Anyway: excellent and fair proposal and excellent work!

dex4er:
Maybe it should be heuristic for the default value per different OS like ie. for node shell. Something is better than nothing then I left just the single value per cluster to set.

PR #165, opened 2025-02-20
The "Remove Hotbar" menu does not appear when more than one hotbar is present

### LINKED ISSUE #163: The "Remove Hotbar" menu does not appear when more than one hotbar is present
**Describe the bug**
The "Remove Hotbar" menu appears only when a single hotbar is present. Creating a new hotbar causes the menu to disappear, making it impossible to delete them.

Furthermore, attempting to delete the only remaining hotbar results in an error: "Cannot remove the last hotbar."

It's as if this feature works the opposite way it should.

**To Reproduce the error on deletion**
1. Click on hotbars menu

Image

2. Click on "Remove hotbar";

Image

3. Click on the only existing hotbar;
4. Click "Remove Hotbar" on the popup;
5. An error will show

Image

**Expected behavior**
When only one hotbar is present, the menu option must not be shown

**To Reproduce the missing menu when more than one hotbar exists**
1. Click on hotbars menu

Image

2. Click "Add hotbar";
3. Write a name and press "enter";
4. Click on hotbars menu again;
5. The "Remove hotbar" menu does not shows up

Image

**Expected behavior**
When more than one hotbar is present, the menu option must be shown

**Environment (please complete the following information):**
- Freelens Version: 1.0.0
- OS: Window, MacOS, Linux
- Installation method: Installer from github

mariomamo:
Fix #163


dex4er (approved):

robertobandini (commented):
@mariomamo Does it matter that these tests fail?
https://github.com/freelensapp/freelens/actions/runs/13446896662/job/37574086884?pr=165

mariomamo:
Hi @robertobandini it has to be fixed but I did not have time yesterday, I will fix it this evening. The problem is about the snapshot of the page, some class names are different, not a big deal

robertobandini:
> Hi @robertobandini it has to be fixed but I did not have time yesterday, I will fix it this evening. The problem is about the snapshot of the page, some class names are different, not a big deal

Thanks, so after you fixed it we can merge it.

dex4er:
Does it fail after `npm run test:unit:updatesnapshot` or you didn't use it yet?

mariomamo:
The test was added with this commit, so the snapshot is "new." The issue occurred because I was working on a different task and had different library versions installed, which caused the generated snapshot to differ from what the pipeline expected. I reinstalled all the correct versions and retried

robertobandini (approved):
Great! Thanks! @mariomamo

PR #167, opened 2025-02-21
lerna@8

dex4er:
More manual updates. This time all devDependencies from top directory. Highlight: lerna v8


dex4er:
It is like Jenga tower. Upgrading this component crashed all others, so I think I will go more complicated way, by enabling Renovate bot for all NPM packages.

PR #176, opened 2025-02-22
Removed SWC and NX from the project; Lerna is used as arbitrary command

dex4er:
Changes:

* Removed SWC from the project. Often it shown some errors that was not presented by tsc and vice-versa. Side effect: tests are slightly slower now: 14m vs 10m.
* @swc/jest was replaced with ts-jest. @swc/cli with webpack or tsc.
* webpack is still in use and has disabled minification. It helps troubleshooting much.
* node-fetch is not repacked with webpack as it was broken without SWC and not compatible with CommonJS anymore.
* Explicit target and modules is "ES2022" however webpack still transpliles to CommonJS.
* Lerna removed from dependencies. Still we use it as `npx -y lerna@6` but it doesn't install hundreds of dependencies in lock file
* NX removed from the project to simplify it, so explicit build is ran before test:unit. Side effect: it adds extra 4m for the workflow, so it is really 20m vs 12m total time.
* Jest configs are in seprate files rather than in package.json.
* Some tests needed extra try/catch, especially in the beforeEach.
* Tests for node-pod extension breaks so some of them are skipped.
* More fixes for typing errors.
* Some tests skipped as they're flaky or fails now without known reason.


dex4er:
Finally 🟢

dex4er:
@mariomamo Please verify. Comments are welcomed.

packages/core/src/common/k8s-api/__tests__/kube-api-version-detection.test.ts
```diff
@@ -30,41 +30,52 @@ describe("KubeApi", () => {
...
-    apiManager = di.inject(apiManagerInjectable);
+      apiManager = di.inject(apiManagerInjectable);
 
-    const setupAutoRegistration = di.inject(setupAutoRegistrationInjectable);
+      const setupAutoRegistration = di.inject(setupAutoRegistrationInjectable);
 
-    setupAutoRegistration.run();
+      setupAutoRegistration.run();
+    } catch (e) {
+      console.error(e);
```
mariomamo, line 55: Why is this catch needed? Is it meant to log the reason for the failure? Was it unclear?

packages/core/src/features/pod-logs/download-logs.test.tsx
```diff
@@ -105,7 +105,8 @@ describe("download logs options in logs dock tab", () => {
       dockStore.closeTab("terminal");
     });
 
-    describe("when logs not available", () => {
+    // TODO: TestingLibraryElementError: Unable to find an element by: [data-testid="download-all-logs"]
```
mariomamo, line 108: Is this activity tracked? Let's make an issue so it doesn't remain pending

packages/core/src/features/pod-logs/download-logs.test.tsx
```diff
@@ -129,7 +130,8 @@ describe("download logs options in logs dock tab", () => {
       });
     });
 
-    describe("when logs available", () => {
+    // TODO: TestingLibraryElementError: Unable to find an element by: [data-testid="download-all-logs"]
```
mariomamo, line 133: Same here

packages/core/src/main/get-metrics.injectable.ts
```diff
@@ -4,7 +4,7 @@
  */
 import { getInjectable } from "@ogre-tools/injectable";
 import type { Cluster } from "../common/cluster/cluster";
-import { FormData } from "@freelensapp/node-fetch";
```
mariomamo, line 7: I see a usage of FormData on line 27. Why has this import been removed?

packages/core/src/renderer/components/catalog/__tests__/custom-views.test.ts
```diff
@@ -14,9 +14,15 @@ import customCategoryViewsInjectable from "../custom-views.injectable";
 
 describe("Custom Category Views", () => {
   let di: DiContainer;
+  // console.log("describe");
```
mariomamo, line 17: Is this a leftover from testing?

packages/core/src/renderer/components/extensions/__tests__/extensions.test.tsx
```diff
@@ -39,42 +39,47 @@ describe("Extensions", () => {
...
+        absolutePath: "/absolute/path",
+        manifestPath: "/symlinked/path/package.json",
+        isBundled: false,
+        isEnabled: true,
+        isCompatible: true,
+      });
+
+      extensionDiscovery.uninstallExtension = jest.fn(() => Promise.resolve());
+    } catch (e) {
+      console.error(e);
```
mariomamo, line 80: Same as already mentioned in other tests

mariomamo (approved):
Hi @dex4er, I left some comments on a few files, nothing serious just some things I noticed. Please take a look, but you did a great job!

packages/core/src/common/k8s-api/__tests__/kube-api-version-detection.test.ts
dex4er: Without this try/catch block if the error occures in this block the test fails without any single explanation. Now it is not so important, as the bug is fixed but still I prefer _any_ single hint what happened rater than only the title of failed test without any explanation.

packages/core/src/features/pod-logs/download-logs.test.tsx
dex4er: It is not tracked yet as I can't imagine we create bugs for PRs :) But yes, it'll be regression here. I think it is acceptable as the functionality still works but the test somehow fails.

packages/core/src/features/pod-logs/download-logs.test.tsx
dex4er: Same here

packages/core/src/main/get-metrics.injectable.ts
dex4er: It is not exported by node-fetch then can't be imported and it is not important as many of such classes are in the lib.dom.d.ts; see https://developer.mozilla.org/en-US/docs/Web/API/FormData

packages/core/src/renderer/components/catalog/__tests__/custom-views.test.ts
dex4er: Good catch. No idea. Probably added automatically by vscode for god-knows-why reason.

packages/core/src/renderer/components/catalog/__tests__/custom-views.test.ts
```diff
@@ -14,9 +14,15 @@ import customCategoryViewsInjectable from "../custom-views.injectable";
...
 
   beforeEach(() => {
-    di = getDiForUnitTesting();
+    try {
+      di = getDiForUnitTesting();
+    } catch (e) {
+      console.error(e);
```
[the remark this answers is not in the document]
dex4er: Another place that errored without any single line of message.

packages/core/src/renderer/components/extensions/__tests__/extensions.test.tsx
dex4er: Same

packages/core/src/renderer/components/catalog/__tests__/custom-views.test.ts
dex4er: Removed

packages/core/src/features/pod-logs/download-logs.test.tsx
dex4er: Additional explanation: TODO here marks additional issue will be created after merge.

dex4er:
BTW one test if flacky. Probably it might happen when Jest starts more in parallel then maybe there is some timeout sometimes:

```
Summary of all failing tests
 FAIL  src/features/cluster/refresh-accessibility-technical.test.ts (18.345 s)
  ● Refresh Cluster Accessibility Technical Tests › given a cluster with no configured preferences › when kubeAuthProxy has started running and its port is found › requests if cluster has admin permissions

    expect(jest.fn()).toBeCalledWith(...expected)

    Expected: anyObject()

    Number of calls: 0

      125 |
      126 |       it("requests if cluster has admin permissions", async () => {
    > 127 |         expect(createSelfSubjectAccessReviewMock).toBeCalledWith(anyObject({
          |                                                   ^
      128 |           spec: {
      129 |             namespace: "kube-system",
      130 |             resource: "*",

      at Object.<anonymous> (src/features/cluster/refresh-accessibility-technical.test.ts:127:51)


Test Suites: 1 failed, 1 skipped, 230 passed, 231 of 232 total
Tests:       1 failed, 134 skipped, 2274 passed, 2409 total
Snapshots:   396 passed, 396 total
Time:        582.794 s
```

packages/core/src/common/k8s-api/__tests__/kube-api-version-detection.test.ts
mariomamo: I imagined it, I agree

packages/core/src/main/get-metrics.injectable.ts
dex4er: Additionally I removed import of unnecessary interface and (re)casting it as everything is fine with this class from lib.dom.d.ts.

mariomamo (approved):

PR #181, opened 2025-02-23
Add CPU and Memory columns

b17k0:
pic


robertobandini:
Hi @b17k0 and thank you for your contribute!

Actually I see only the unit tests failing.
Additional columns can still be enabled or disabled like the others, right?

Thanks

b17k0:
Yes, functionally everything works as expected. However, I rushed a bit with the PR. I still need to fix the unit tests, and I can already see some areas for optimization. Unfortunately, I’ll only have time to work on this over the weekend.

Also, my TypeScript coding skills are far from perfect. There are some convenient features in the proprietary version that are missing in freelens, and this is just an attempt by a non-professional to bring something I’d love to have in freelens.

If someone with more TS experience is interested, we could do this together. I’m sure the result would be much better that way!

robertobandini:
Hi @b17k0 thank you!
I already added some reviewers.
@mariomamo do you think you can provide some help/review here?

mariomamo:
Yes of course, it already was on my list, I'll take a look tonight

mariomamo (changes requested):
1) Your classes contain too many elements. I suggest creating separate files for the non-injectable component (`NonInjectablePodMetricsMemoryContent`) and the injectable one (`InjectablePodMetricsMemoryContent`). `NonInjectablePodMetricsMemoryContent` should only contain the component logic: HTML elements, functions, variables, and so on, while `InjectablePodMetricsMemoryContent` should only return the component wrapped in `getInjectable`, as you already did. You can take `pod-logs-menu.tsx` and `pod-logs-menu.injectable.ts` as examples.  

2) Some functions are duplicated: `calculatePodKubeMetrics` and `getPodKubeMetrics`. Could you refactor them into a separate file and call them where needed? I suggest using a hook that exposes these two functions, something you can use like this:  

   ```typescript
   const kubeMetrics = useKubeMetrics();
   kubeMetrics.getPodKubeMetrics();
   ```

3) The functions `getPodKubeMetricsMemoryDisplayValue` and `getPodKubeMetricsCpuDisplayValue` are declared outside the `PodMetricsMemoryContent` and `PodMetricsCpuContent` declaration, respectively. I would move these functions inside them.  

4) Could you please add a unit test to verify that everything is working correctly? Unit tests are very useful to prevent regressions when someone else updates the code.  

These might seem like many points, but they are just small improvements that can make the code more readable, in my opinion 😄. Please reach me out if you need help or if you disagree with something, and we can discuss it.

mariomamo:
I runned the code and the app does not crash if a metric server is missing
image

maifeeulasad:
I think this one can be closed, as this feature was already handled via a different merge request.

PR #183, opened 2025-02-24
Node 20 eslint plugin 7 update

### LINKED ISSUE #161: Update Node to version 20 and eslint-plugin to version 7 in the root package.json
Updating the Node version to 20 in the root package.json file requires upgrading TypeScript to version 7. These changes fix 30 vulnerabilities. This update involves changes to object types and the addition of type annotations in 90 files, which need to be thoroughly tested

mariomamo:
This resolve #161


robertobandini:
Hi @mariomamo I see the tests failing.
Is this connected with https://github.com/freelensapp/freelens/pull/176 ?
You and @dex4er have you already talked about this?
Thanks

mariomamo:
Yes, @robertobandini, I found that there are many lint and test issues that I'm currently fixing. In this commit, I only updated Node to version 20 in the root project folder and upgraded ESLint to version 7.

@dex4er, have you already made any of these updates?

dex4er:
@mariomamo Yes, I think my changes might help you with your upgrades.

I'm trying to reduce the compilers as much as possible. This stack is hilarious: it uses SWC, tsc and webpack and in the same time.

Now we have the situation that different syntax (variant of Typescript) is used for tests and by application. As all we use standard plugins for VScode then it is better to stick with tsc as default. Still, unfortunately, tsc in ESM mode is transpiled to commonjs and it makes a lot of subtle problems. The most problematic is node-fetch which is not compatible with commonjs anymore.

I don't feel so brave to switch to ESM modules without webpack, but sooner or later we'll have to do it. After I finish my PR #176 it should be easier to troubleshoot tests as they'll use the same compiler as all other parts.

mariomamo:
Seems like the integration test for Windows is stuck and ends after one hour due to a timeout. It says "Ran all test suites," but nothing happens.
image

mariomamo:
When I started this PR, it fixed about 30 vulnerabilities. Now, after all the PRs made in the past few days, there are only 15 vulnerabilities left to fix.

@dex4er, can you please have a look?

Issue #434, opened 2025-03-19
Can't paste text into "Add Cluster" text area


a563151:
### Describe the bug

Can't paste text into "Add Cluster" text area.

*Also, dragging text from another text area brings the text without newlines, regardless of Unix / Windows new lines.*

### Reproduction steps

1. Go to "File -> Add Cluster"
2. Paste text copied from other text areas
3. See no change in the "Add Cluster" text area

### Expected behavior

See text from clipboard in the text area

### Screenshots

[Window Title]
  Freelens

[Main Instruction]
Freelens

[Content]
Freelens: 1.1.0
Extension API: 1.1.0
Electron: 34.3.3
Chrome: 132.0.6834.210
Node: 20.18.3
Platform: win32
Architecture: x64
© 2024-2025 Freelens Authors

[Close]

### Freelens version

1.1.0

### Operating system type and version

Windows 11 Enterprise

### Installation method

Freelens-1.1.0-windows-amd64.exe

### Logs

(electron) 'webContents.canGoBack' is deprecated and will be removed. Please use 'webContents.navigationHistory.canGoBack' instead.
(electron) 'webContents.canGoForward' is deprecated and will be removed. Please use 'webContents.navigationHistory.canGoForward' instead.

damiendesvent:
Have you tried using the CTRL + Shift + V shortcut instead of the CTRL +V?
For me I have to press the Shift key for copy-paste, in the _Add cluster_ section and even in other section like shells or manifest editors.
Though, for me it was already something needed in Lens.

a563151:
CTRL + SHIFT + V does work for me too, I can at least use that as a workaround, thanks. CTRL + V or "Right Click" -> "Paste" does not work.

albundy83:
It seems that it's CTRL +V which does not work but in the entire application, not just in the "Add Cluster" area :)
Middle click button for the linux folks works by the way

dex4er:
Fixed in https://github.com/freelensapp/freelens/pull/446 and will be released with v1.1.1

PR #446, opened 2025-03-21
Fix for problem that paste operation doesn't work in an editor

### LINKED ISSUE #445: When editing resource cant copy/paste
### Problem

you can not copy and paste resource edit section. right click + paste not working, also ctrl+v not working. 

### Version

v1.1.0

dex4er:
Fixes #445, #434

**Description of changes:**

- Added security context for iframe that enabled read-write operations on clipboard
- Patch for VS Code (monaco-editor) which enables alternative clipboard handler on newer Chrome

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "postdev": "npx -y lerna watch -- npx -y lerna run build --stream --include-dependents --scope \\$LERNA_PACKAGE_NAME",
     "prestart-dev": "cd freelens && npm run build:tray-icons && npm run download:binaries",
     "start-dev": "npx -y lerna run start",
-    "postinstall": "linkable",
+    "postinstall": "linkable && npx -y patch-package@8",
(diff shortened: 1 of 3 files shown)


dex4er:
@mariomamo This is the workaround without fully understanding what is the reason why we cannot paste it from the clipboard now. Do you have any ideas?

Still, pasting using context menu also doesn't work.

nmengual:
Maybe the problem is related to this:
https://github.com/microsoft/monaco-editor/issues/4402

dex4er:
Ok, so it is possible to add own action: https://microsoft.github.io/monaco-editor/playground.html?source=v0.52.2#example-interacting-with-the-editor-adding-an-action-to-an-editor-instance

mariomamo:
Hi @dex4er I just read your comments but I see you already fixed it 😊, great work!

dex4er:
@mariomamo Yes, it was a bug in VS Code to fix.

PR #448, opened 2025-03-21
Remove text selection style causing readability issues in light theme

nmengual:
Removed custom text selection styling that was causing readability issues in the light theme, where selected text would become invisible (white text on white background). This change follows the behavior already implemented in the Lens application.


dex4er:
Thanks a lot. It really helps!

PR #449, opened 2025-03-21
Update Electron to 35.0.3 and Node.js to 22.14.0

nmengual:
This PR updates our application dependencies to use the latest stable versions:

Upgraded Electron from previous version to 35.0.3
Updated Node.js to 22.14.0

This resolves the issue preventing React DevTools from working properly, enabling better debugging and development experience


dex4er:
Oh great! Thanks for your support as we have some dependencies waiting Node 22.

However it does break with

```
  npm error engine Not compatible with your version of node/npm: @freelensapp/core@1.1.0
  npm error notsup Not compatible with your version of node/npm: @freelensapp/core@1.1.0
  npm error notsup Required: {"node":"^20.0.0"}
  npm error notsup Actual:   {"npm":"10.9.2","node":"v22.14.0"}
```

busches:
Should upgrade `@types/node` as well.

nmengual:
I have fixed the problem with the @freelensapp/core@1.1.0 package (using --force in the install, which I'm not sure is the correct option) and updated the @types/node dependency.

nmengual:
It seems that the problems in that test are with the version of node (with 20 they work, with 22 they don't) I'm investigating

nmengual:
I have manually tested the functionality that fails in the tests (managing Helm repositories) and it seems to be working correctly so I have opted to update the snapshots for that part.

nmengual:
I think it's ready

packages/core/src/features/helm-charts/__snapshots__/add-custom-helm-repository-in-preferences.test.ts.snap
```diff
@@ -1100,40 +1100,12 @@ exports[`add custom helm repository in preferences when navigating to preference
                       Synced Items
                        
                     </div>
-                    <div>
+                    <div
```
busches, line 1103: This seems like a mistake and the tests need to be updated to wait for the loading to finish to assert on the data, instead of asserting on the spinner.

nmengual: Do you know which part of the test I should wait for? I've tried several sites and have the same problem. The results are different depending on whether you run it with Node 20 or 22.

nmengual: I was able to solve it by waiting for the component to be painted and I only had to update a snapshot that shows a class (leave)

dex4er:
I still would like to see it in next v1.2.0 rather than v1.1.1 today. Electron 35 and Node 22 will bring more things to test. Even VS code doesn't support it yet then ie. Electron 34 broke Monaco Editor and copy-paste.

I'll try to not break this PR with more updates. Thanks for your patience and rebases!

dex4er:
@nmengual One more rebase please. I guess the conflict is after the version change.

dex4er:
Ah forget it, there will be v1.1.2 too

nmengual:
I already updated the branch

dex4er:
@nmengual Many thanks!

PR #453, opened 2025-03-22
Replace cross-env with x-var and rimraf with shx

dex4er:
Not so important but cross-env looks to be inactive and shx and x-var are active. Eventually running them with npx also might be a good option.


dex4er:
So `x-var` can't replace `cross-env`. Hopefully `cross-env` will be updated from time to time.

PR #481, opened 2025-03-25
Fix spacing issue in error boundary

busches:
There was no space between these words and the link.

Didn't see any tests for this file, but it fixed it locally.


dex4er:
Thanks for fixing error in presenting errors! :)

PR #482, opened 2025-03-25
Upgrade to MUI 5

busches:
Lab .142 is the last version that actually has `TreeView`, otherwise it wants to install MUI 6 and renders nothing. `TreeView` had breaking changes and to get only the icon to do the expansion like current state, it required that fun new component.

Also upgraded react-material-ui-carousel for MUI 5, but I didn't see an easy way to confirm this still works, there were no tests and I don't have a carousel locally on the welcome page. Same goes for the `LargeTooltip`.

Once this is merged, we'll be able to upgrade to React 18 now.


dex4er:
Thanks a lot! 

I think first we'll merge it then we'll find a way to further updates.

I would remove carousel. I think it is for advertising extensions but I really don't care about it: welcome page might be really simple.

PR #507, opened 2025-03-28
Removed carousel with welcome banners

dex4er:
It is actually breaking change for Freelens extensions but I don't think any existing extensions adds own banners yet, so I don't plan to bump our extension API.

The intention of this removal is to reduce dependencies on the React components that are already dead projects. `react-material-ui-carousel` is not updated for 3 years and stuck on MUI 5.


dex4er:
@busches I guess it will help you?

dex4er:
Sure we'll try! Maybe not today, but I don't see any point to stick with MUI 5 forever.

PR #555, opened 2025-04-09
Migrate to Pnpm from NPM with Lerna

dex4er:
The new project setup that replaces NPM and Lerna.

Pnpm uses package isolation then a lot of dependencies are fixed. Also all peerDependencies has been replaced with dependencies or devDependencies as they are now handled correctly by the package manager (with the exception for jest, webpack and its loaders that are hoisted to the top package).

Pnpm improves our workflows and now builds and tests are much faster!

Devserver is (still) broken and probably will be finally fixed after migrating to Vite.


dex4er:
I'm not quite understand why test units fails and reports new snapshots.

dex4er:
It runs integration tests on Mac only for 4m 32s? Wow.

patches/app-builder-lib.patch
mariomamo, line 1: Should this patch be removed?

patches/monaco-editor.patch
mariomamo, line 1: Should this patch be removed?

PR #556, opened 2025-04-10
docs: add homebrew tap command

breuerfelix:


dex4er:
Hi. It is not needed anymore as now Freelens is a part of official Homebrew Cask project:

https://formulae.brew.sh/cask/freelens

PR #608, opened 2025-04-12
Fix for Powershell on Mac - Closes #480

### LINKED ISSUE #480: The terminal not work as expected when using pwsh on mac
### Describe the bug

I use the `powershell` for my default shell on Mac arm64, when tried to execute a command  in the pod, could not into the pod terminal, when removed the `exec` from the command, it worked.
And when I changed the terminal to `bash` or `zsh` it also works

### Reproduction steps

screen recording:

https://github.com/user-attachments/assets/f954fb2a-6618-4bbf-bde0-11655dad57f4

### Expected behavior

It should work for pwsh also

### Screenshots

_No response_

### Freelens version

1.1.1

MacOS 15.3.2 M3

### Installation method

brew install

### Logs

_No response_

### Kubeconfig

_No response_

### Additional context

_No response_

camclay:
Draft MR to handle the various places where Shell logic is parsed, and consolidate it so it's checked on initialization or against the User set preference shell, and ensure the logic works consistently.


robertobandini:
Thanks about this @camclay!

camclay:
Think it's worth merging this, after some clean up of comments, and then tackling the larger refactor separately as the main fix, is very small.

camclay:
Cleaned up, and ready for additional testing!
Feature test can be seen here - https://drive.google.com/file/d/1tGWs67UzleAzE-FVfDvYFfeb1s_RM5gK/view?usp=sharing

robertobandini:
> Cleaned up, and ready for additional testing! Feature test can be seen here - https://drive.google.com/file/d/1tGWs67UzleAzE-FVfDvYFfeb1s_RM5gK/view?usp=sharing

Really appreciated! Thanks!

packages/core/src/renderer/components/node-pod-menu/pod-attach-menu.tsx
mariomamo, line 1: Duplicated comment here, can you remove one of them?
Are the commented imports necessary? If not, can they be removed as well?
Same for the other comments

packages/core/src/renderer/components/node-pod-menu/pod-shell-menu.tsx
mariomamo, line 1: Are the commented imports necessary? If not, can they be removed?
Same goes for other commented-out code.
If the TODO hasn't been addressed yet, please create an issue for it so we don’t forget.

mariomamo (approved):
I really appreciate the documentation you've written in Markdown format, great work 🙂

robertobandini:
Thanks about your help @mariomamo!

camclay:
Heya, will work on the additional commented items and tests this evening, will advise if I hit any roadblocks! Thanks!

dex4er:
@camclay Hi. Beside my vacation 🏜️ I was focused on finishing the internal refactoring and replacing ESLint with Biome. It caused reformatting of all files. Is it any chance you can rebase your PR on the latest main?

camclay:
Will do! Hoping to wrap this up today, and can rebase for sure!

packages/core/src/renderer/components/node-pod-menu/pod-shell-menu.tsx
camclay: Removed! Thanks for catching those!

packages/core/src/renderer/components/node-pod-menu/pod-attach-menu.tsx
camclay: Removed!

dex4er:
So the fix is incorrect because after entering the pod shell then exiting it the tab is still opened then it needs an explicit `exit` command or closing the tab.

This is the reason why `exec` was originally added as the first argument.

dex4er (changes requested):
`commandParts.unshift("exec");` is removed and there is no different logic for it then finally on Linux and Mac with POSIX-like shell (bash, zsh) the tab is not autoclosed.

dex4er:
Hmmm now I think it is reasonable to leave the terminal after any command finishes. @robertobandini what do you think? Maybe we could introduce such a change. We could also append `&& exit` instead `exec` at the beginning and it would be more portable.

camclay:
I'm open to segmenting the logic, and/or making it more portable. This is a weird one, as it seems to have edges on all platforms!

dex4er:
Ok I found that the universal syntax is `command ; exit` and it is supported by POSIX shells, fish and Powershell.

Also another patch is on the way that adds "This terminal will be auto-closed in 15 seconds" then delays closing the tab with the Terminal. In that case this `; exit` can be added safely to every launched command, ie. for `kubectl drain` or `kubectl cordon` too and we can replace this `exec` with `... ; exit` for others.

dex4er:
https://github.com/freelensapp/freelens/pull/748 sets the delayed auto-close for any type of command so it should be save to add `; exit` to all our commands in the terminal

dex4er:
Aha, so the implementation of running kubectl commands here is just incorrect. Currently: The standard terminal with shell is opened then kubectl command is injected into terminal as it would be a keystroke.

This is just wrong.

Ofc it should be started "kubectl" binary with the proper parameters without even running the standard shell.

I'm looking into code how simple would be to replace this behavior for kubectl attach / exec / cordon / uncordon.

BTW, Node shell feature is already correctly implemented and it really works without spawning the system shell.

robertobandini:
> Aha, so the implementation of running kubectl commands here is just incorrect. Currently: The standard terminal with shell is opened then kubectl command is injected into terminal as it would be a keystroke.
> 
> This is just wrong.
> 
> Ofc it should be started "kubectl" binary with the proper parameters without even running the standard shell.
> 
> I'm looking into code how simple would be to replace this behavior for kubectl attach / exec / cordon / uncordon.
> 
> BTW, Node shell feature is already correctly implemented and it really works without spawning the system shell.

Sorry I'm really behind with notifications, I reread now and for what you described I definitely agree too.

PR #654, opened 2025-04-25
fix: update `notarize.js` to use dynamic import for `@electron/notarize` module

omarluq:
Replace CommonJS `require()` with ES Module dynamic import to resolve `require() of ES Module not supported` error during build process. The `@electron/notarize` package is now an ES Module and requires this approach for compatibility with electron-builder.


dex4er:
I'm not sure if we need/can this patch but let's try. We'll figure it out tomorrow after nightly build.

PR #745, opened 2025-05-21
build: replace cross-env with pnpm shell-emulator

arnaud-dezandee:
Replace `cross-env` with built-in [pnpm shellEmulator](https://pnpm.io/cli/run#shellemulator)


dex4er:
@arnaud-dezandee Good idea! Thanks for patch.

PR #775, opened 2025-05-24
Use hub.helm.sh to fetch the list of Helm repos

dex4er:
This is the same URL that is used by `helm search hub` command.


arnaud-dezandee:
Nicely found! Should we use the helm-managed redirection if they ever wish to change the default?
[`https://hub.helm.sh`](https://github.com/helm/helm/blob/main/pkg/cmd/search_hub.go#L73)

PR #776, opened 2025-05-25
Add CPU and Memory usage metric column for pods view

### LINKED ISSUE #743: CPU/Memory metrics on Pods list

Just like in Lens, I’d like to see CPU and Memory metrics displayed directly on the Pods list view.

Lens
Image

FreeLens (currently no metrics on the list; only visible when selecting individual pods)
Image

### Why is this needed?

Having CPU/Memory metrics directly on the list view gives a quick overview of resource usage across all pods, whether scoped to a single namespace or across all namespaces. It helps in identifying high-usage pods faster without needing to inspect them one by one.

HanCheo:
I saw PR #181, but it hasn’t been updated since then.
So I’ve created a new one.

Screenshot of the current state:
image

Here’s what it looks like when the metrics-server isn't working:
image

this will close  #743

### DIFF
--- a/freelens/build/metainfo.xml
+++ b/freelens/build/metainfo.xml
@@ -69,7 +69,7 @@
   <icon type="remote">https://freelensapp.github.io/images/freelens-logo-512.png</icon>
   <content_rating type="oars-1.1"></content_rating>
   <releases>
-    <release version="1.3.0" date="2025-05-09">
+    <release version="1.3.1-0" date="2025-05-25">
       <description translatable="no">The latest release.</description>
     </release>
   </releases>
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -4,7 +4,7 @@
   "productName": "Freelens",
   "description": "Freelens - Free IDE for Kubernetes",
   "homepage": "https://freelens.app",
-  "version": "1.3.0",
+  "version": "1.3.1-0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/freelensapp/freelens.git"
--- a/packages/business-features/keyboard-shortcuts/package.json
+++ b/packages/business-features/keyboard-shortcuts/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@freelensapp/keyboard-shortcuts",
   "private": false,
-  "version": "1.3.0",
+  "version": "1.3.1-0",
   "description": "Keyboard shortcuts for Freelens",
   "type": "commonjs",
   "files": [
--- a/packages/cluster-settings/package.json
+++ b/packages/cluster-settings/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@freelensapp/cluster-settings",
-  "version": "1.3.0",
+  "version": "1.3.1-0",
   "description": "Injection token exporter for cluster settings configuration",
   "license": "MIT",
   "type": "commonjs",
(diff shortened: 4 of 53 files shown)


packages/core/src/renderer/components/workloads-pods/columns/pods-cpu-usage-column.injectable.tsx
```diff
@@ -0,0 +1,38 @@
...
+    id: columnId,
+    kind: "Pod",
+    apiVersion: "v1",
+    priority: COLUMN_PRIORITY.CPU_USAGE,
+    content: (pod) => {
+      const podStore = di.inject(podStoreInjectable);
+      const metrics = podStore.getPodKubeMetrics(pod);
+      const cpuUsage = metrics.cpu
+      
+      return <span>{cpuUsage === 0 ? "N/A" : cpuUsage.toFixed(3)}</span>;
```
HanCheo, line 27: I'm wondering whether it's appropriate to show N/A when the CPU value is 0, given that 0 is the default value.
Wouldn't it make more sense to show N/A only when the metric server is unavailable or the data can't be retrieved?

For reference, memory currently displays N/A when the value is 0, so this change was made for consistency.

HanCheo, line 27: I'm wondering whether it's appropriate to show N/A when the CPU value is 0, given that 0 is the default value.
Wouldn't it make more sense to show N/A only when the metric server is unavailable or the data can't be retrieved?

For reference, memory currently displays N/A when the value is 0, so this change was made for consistency.

arnaud-dezandee: Hi @HanCheo, I'd expect `N/A` to be shown only when the metrics server is truly unavailable. Otherwise, we should print out the value returned by it, whatever it may be

HanCheo: Yep I'm really agree. 
but... currently, getPodKubeMetrics returns { cpu: 0, memory: 0 } when no metrics are found.
However, returning 0 can be misleading, as it implies that the metrics exist and the usage is simply zero.
It would be more appropriate to return something like "N/A" to explicitly indicate that no data is available.
This helps distinguish between "no usage" and "no metrics."

And dosen't need compare cpuUsage === 0 

```ts
  getPodKubeMetrics(pod: Pod) {
    const containers = pod.getContainers();
    const empty = { cpu: 0, memory: 0 };  // <<-- here
    const metrics = this.kubeMetrics.find((metric) => {
      return [metric.getName() === pod.getName(), metric.getNs() === pod.getNs()].every((v) => v);
    });

    if (!metrics) return empty;
```

HanCheo: And Memory ByteToUnits code. return "N/A" when <= 0.

```
export function bytesToUnits(bytes: number, { precision = 1 }: BytesToUnitesOptions = {}): string {
  if (bytes <= 0 || isNaN(bytes) || !isFinite(bytes)) {
    return "N/A";
  }
```

HanCheo: I change it!

dex4er:
Does it really work? I have EKS cluster and I see N/A:

image

HanCheo:
@arnaud-dezandee 
I think returning NaN is causing a wrong sort.
I will fix it

HanCheo:
@dex4er yes!

PR #835, opened 2025-06-08
YAML dump with consistent formatting

dex4er:
Following parameters are used everywhere:

```json
{
    noArrayIndent: true,
    noCompatMode: true,
    noRefs: true,
    quotingType: '"',
    sortKeys: true
}
```

### DIFF
--- a/packages/core/src/common/k8s/resource-stack.ts
+++ b/packages/core/src/common/k8s/resource-stack.ts
@@ -132,7 +132,9 @@ export class ResourceStack {
           labels["app.kubernetes.io/created-by"] = "resource-stack";
         }
 
-        resources.push(yaml.dump(entry));
+        resources.push(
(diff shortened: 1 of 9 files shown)


packages/core/src/common/k8s/resource-stack.ts
```diff
@@ -132,7 +132,9 @@ export class ResourceStack {
           labels["app.kubernetes.io/created-by"] = "resource-stack";
         }
 
-        resources.push(yaml.dump(entry));
+        resources.push(
+          yaml.dump(entry, { noArrayIndent: true, noCompatMode: true, noRefs: true, quotingType: '"', sortKeys: true }),
```
busches, line 136: Would it make sense to move this shared config to a central place or wrap this function so it's always consistent?

PR #937, opened 2025-06-30
Prometheus graphs ignores metrics with empty container label

RubenNL:
freelens doesn't calculate the pod cpu, memory and filesystem usage correctly. Here's an example of a single-container pod:

Before:
before-pod
before-container

After:

after-pod
after-container

I have this issue with those metrics: 

```
container_cpu_usage_seconds_total
container_memory_working_set_bytes
container_fs_usage_bytes
container_fs_writes_bytes_total
container_fs_reads_bytes_total
```

Those are all metrics where the name starts with `container_*`, excluding `container_network_*` (Those never have a `container` attribute).

All other `*-provider.injectable.ts` have this same fix applied. I haven't been able to find why this one is different, so let's set them to the same queries.

**Info about my cluster**

I'm running k3s, with one master node and multiple "worker" nodes.  
Prometheus is provided by the `bitnami/kube-prometheus` helm chart.

**Tests**

The tests have successfully ran on a testing PR on my own repo: https://github.com/RubenNL/freelens/pull/1

### DIFF
--- a/packages/technical-features/prometheus/src/operator-provider.injectable.ts.ts
+++ b/packages/technical-features/prometheus/src/operator-provider.injectable.ts.ts
@@ -82,23 +82,23 @@ export const getOperatorLikeQueryFor =
       case "pods":
         switch (queryName) {
           case "cpuUsage":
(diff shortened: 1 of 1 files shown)


robertobandini:
Thanks @RubenNL!

PR #958, opened 2025-07-06
Add ephemeral containers to pods view

### LINKED ISSUE #184: Add support for debug/ephemeral containers
**What would you like to be added**:

Like the remote shell functionallity in the pod menu I would like to have the abillity to run [debug](https://kubernetes.io/docs/tasks/debug/debug-application/debug-running-pod/#ephemeral-container)/[ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/).

**Why is this needed**:

This is required for pods where a simple remote shell does not work, e.g. when the container runs with distroless images or without debugging tools.

**Environment you are Lens application on:**

- Kubernetes distribution: k8s
- Desktop OS: MacOS

dex4er:
**Description of changes:**

Closes #184

- [x] Add ephemeralContainers to Pods API
- [x] Add to pod details
- [ ] Start a new ephemeral container with the image/command from settings (and/or parameters from dialog)
- [ ] Terminate ephemeral container

Terminating the ephemeral container is tricky as Kubernetes does not hold information about the main PID for containers with shared PIDs. The workaround might be either adding some information about main PID to logs or using cgroups to find out the main PID to kill.

Creating/killing the containers will be in a separate PR.

### DIFF
--- a/packages/core/src/renderer/components/__tests__/cronjob.store.test.ts
+++ b/packages/core/src/renderer/components/__tests__/cronjob.store.test.ts
@@ -36,7 +36,7 @@ const scheduledCronJob = new CronJob({
           metadata: {},
           spec: {
             containers: [],
-            restartPolicy: "restart",
+            restartPolicy: "Always",
             terminationGracePeriodSeconds: 1,
             dnsPolicy: "no",
             hostPID: true,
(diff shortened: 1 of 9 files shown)


victor-elizondo:
Hi @robertobandini 
my team is currently moving towards shell-less / distroless containers for security reasons, so having a clean workflow to launch debug containers is becoming increasingly important for us.

That’s actually how I ended up here. We initially relied on a Lens extension for this capability, and when moving to Freelens I found the Debug extension again. Then I noticed there has been some work towards supporting ephemeral containers directly in the core app, which is awesome.

I was wondering if the UI support for starting ephemeral containers is already available or planned to be included at some point?
It’s also very possible I’m just missing it 🙂

I’d love to have the ability to start a new ephemeral container directly from the UI.

Thank You!

PR #979, opened 2025-07-11
Use tailwind from CSS file rather than SCSS

dex4er:
Using it from SCSS shows errors that directives are not compatible with sass modern mode.


dex4er:
So it doesn't work: probably right now it is just an annoying warning and not much harm, so I'm closing it.

PR #1048, opened 2025-08-04
Include more details for the Cronjobs executions

knives85:
As Freenlens users I felt useful to provide this small contribution regarding the Cronjob details page
In particular here we have the following small changes:
* The list of Cronjob executions is now sorted by descending `startTime`
* Each element in the list is now identified by the job unique name and not using the general Cronjob name
* Each element in the list now shows also the `Start Date` and the `Duration`

Hope these changes could be useful also for other users.


dex4er:
Thank you! It looks really useful.

PR #1059, opened 2025-08-08
Feature: pod force terminate button with terminal invocation

kappa8219:
It adds an ability to terminate pod with forced parameters(see the last screenshot). It is usefull when dealing with pod stuck in "Terminating" state.

Screenshot From 2025-08-08 10-57-25

Screenshot From 2025-08-08 10-57-51

Bye.. Bye...

Screenshot From 2025-08-08 10-58-13


dex4er:
I've found that we have now propagation policy in API

https://github.com/freelensapp/freelens/blob/92e82084444f0cb4efbee25c67360c202a95d64e/packages/utility-features/kube-api/src/kube-api.ts#L729

We can add more options for delete.

PR #1147, opened 2025-08-29
[feature]: force delete pod on terminating, failure, etc;

### LINKED ISSUE #416: Add ability to force pod delete
- You may find yourself needing to forcibly delete a pod
- It is not currently possible via the delete button of a pod on Freelens
- It would therefore be useful to have it and possibly be able to specify the parameters
- For example "kubectl delete pod pod_name --grace-period=0 --force --namespace pod_namespace"

If I am wrong about this not being possible at the moment or for any other comments, help or contributions on this matter, please contact us, thanks!

maifeeulasad:
closes #416

Here is how it works:

https://github.com/user-attachments/assets/87d713ed-246e-4865-aa04-81f8a3913b94

These are the images I tested with:
```
kubectl run nginx-test --image=nginx --namespace=test
kubectl run busybox-test --image=busybox --namespace=test -- sleep 3600
kubectl run alpine-test --image=alpine --namespace=test -- sh -c "while true; do sleep 30; done"
```

**I am open for further review.**

*NB: Please assign me to the task before you close it, after some hard days I love to show off that I have done this or that.*


maifeeulasad:
Do I update the snap for test cases? Or do we take care of that with different MR?

New to this codebase. Can someone tell me, what are the other formalities I need to do?

maifeeulasad:
It's done. I have updated both the snap and linting issues.

maifeeulasad:
Ok, I can see few unit tests are failing. I am looking into them.

I will share my findings here.

dex4er:
I test this implementation.

For freshly started and running pod from CronJob I see Force Delete button. Actually why? I didn't try to delete it yet. I guess just because it has finalizers.

image

Clicking it doesn't terminate immediately:

image

And for this pod

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: stuck-pod
spec:
  terminationGracePeriodSeconds: 30
  containers:
  - name: ignore-sigterm
    image: bash
    command: ["bash", "-c", "--"]
    args:
      - |
        trap '' TERM INT
        echo "Ignoring SIGTERM, waiting forever..."
        while true; do sleep 5; done
```

I see 

image

and clicking it doesn't do anything.

Nice prototype, but it looks like the logic is broken.

Anyway, maybe for pods it is good to add an extra button with "[Force Delete]", then I won't need to click it 2 times for the same pod?

image

Checking the context (if resource still waits for finalizers) is a good idea: showing the same button for deletion that doesn't do anything is not the right way, so after detection that resource is deleted (deletionTimestamp?) I imagine I can see "terminate" button instead of "delete".

I think it is a good starting point, but we didn't get a working solution yet.

The correct specification of the feature would be:

1. To allow normal graceful deletion by default, as we already do.
2. To allow optional force deletion, which means deletion with gracefulShutdown=0 (I guess it makes sense only for pods?). We can present then 2 buttons on the dialog for deletion. Force deletion button should be red rather than cyan.
3. To allow force deletion if pod is currently in a terminating state (deletionTimestamp). Presenting a normal delete button does not make sense anymore in this case.
4. To allow termination (it is better to avoid "deletion" in this context) only when the resource is currently terminated and waits for finalizers.

How to distinguish terminating and finalizing:

Terminating (waiting for grace period):
- metadata.deletionTimestamp set
- status.containerStatuses show running or waiting
- Finalizers may or may not exist

Terminated but not finalized:
- metadata.deletionTimestamp set
- All status.containerStatuses show terminated
- Pod still exists because metadata.finalizers is non-empty

As you can see, killing a pod is a complicated thing :)

Thanks for your contribution!

robertobandini:
Hi @maifeeulasad thank you very much about your contribute!

maifeeulasad:
Thanks @robertobandini ! It's an honor to contribute to this awesome project.

I have fixed the failing test, care to check again please?

maifeeulasad:
@dex4er All the tests are passing now. Care to share your insight on this one please?

Open to additional modifications as well.

dex4er:
For the record (and snippet for testing). the stuck finalizer is:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: stuck-finalizer-pod
  finalizers:
    - example.com/stuck
spec:
  terminationGracePeriodSeconds: 0
  containers:
  - name: ignore-sigterm
    image: bash
    command: ["bash", "-c", "sleep 99999"]
```

dex4er:
Problematic case:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: stuck-pod
  finalizers:
    - example.com/stuck
spec:
  terminationGracePeriodSeconds: 600
  containers:
  - name: ignore-sigterm
    image: bash
    command: ["bash", "-c", "--"]
    args:
      - |
        trap '' TERM INT
        echo "Ignoring SIGTERM, waiting forever..."
        while true; do sleep 5; done
```

it breaks after normal delete.

dex4er:
As I checked API: status.phase is one of `Pending | Running | Succeeded | Failed | Unknown`

Force delete won't work for `Succeeded | Failed | Unknown` then we should skip it for such statuses.

maifeeulasad:
I have seperated logic for deletion based on type and state as well.

And here is the logic summarizing the changes of the last commit:
 - For Pods not yet terminating:
   - Succeeded / Failed / Unknown phases -> only delete option
   - Running / Pending phases -> both `force_delete` and delete options
 - For Pods already terminating:
   - With finalizers -> only `force_finalize` option
   - Succeeded / Failed / Unknown phases -> only delete option
   - Running phase or with running containers -> `force_delete option
   - Otherwise -> delete option
 - For Generic Resources:
   - Not terminating -> delete option
   - Terminating with finalizers -> `force_finalize` option
   - Terminating without finalizers -> delete option
   
   
 Care to check it again please?

dex4er:
`stuck-pod` can't be deleted or terminated so my guess it still does not work correctly.

maifeeulasad:
@dex4er can we get it merged for now and fix these issues as bugs in upcoming days?

I think it will be really helpful as then we will get real world feedback. Almost all cases are working, except for one or two trapping exceptions. I am not even sure if they will be present in a real world scenario or not. Not rejecting them as a valid test case, just want to get the feature rolled out.

dex4er:
I'll try to fix it, but generally I want to avoid releasing an obviously incorrect feature. Actually, the cases are not unique, as I found them on my standard cluster with no fancy setup. I could agree we can have something broken for what was already broken but not any new regression.

dex4er:
Similar test for non-pod resource:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: stuck-ns
  finalizers:
    - example.com/stuck
```

dex4er:
Regression: doubles "delete" button:

image

dex4er:
Ah namespace is very special and has dedicated "delete" button. I think we can remove this very dedicated "delete" menu for namespace and eventually add extra features for global "delete" button in future.

dex4er:
Finally. Many kudos @maifeeulasad It was great cooperation!

Issue #28, opened 2024-10-04
Evaluate if troubleshooting and jsonnet dirs are still needed


robertobandini:
As in the title, I mean, dirs and their files in the root dir of the freelens repo

PR #1235, opened 2025-09-25
Markdownlint silenced

dex4er:
**Description of changes:**

- Correct configuration applied with extended list of ignored problems


robertobandini:
So, given the files changed here I imagine that we can close also this old issue by me isn't it?
https://github.com/freelensapp/freelens/issues/28

PR #1248, opened 2025-09-29
Sorting of left menu items 

### LINKED ISSUE #1205: Sorting of left menu items
### Discussed in https://github.com/freelensapp/freelens/discussions/1149

<div type='discussions-op-text'>

<sup>Originally posted by **skrivopishin** August 29, 2025</sup>
It would be great, if it will be possible to sort items on the left, e.g. put FluxCD above Workloads.
And maybe even hide something.</div>

mariomamo:
#Closes #1205


mariomamo:
@dex4er @robertobandini I finally finished it, now the settings are global for all clusters as we agreed, could you please do a review?

mariomamo:
I can't think about it, the risk is that this UI become too messy

PR #1301, opened 2025-10-29
Add logs button column to pod list view

pezzking:
## Summary

Adds a new "Logs" button column to the pod list view, allowing users to quickly open pod logs with a single click without needing to use the context menu.

## Changes

- Added `pods-logs-button-column.injectable.tsx` with logs button component
- Added logs column priority to `column-priority.ts`

## Features

- Clicking the logs button opens logs for the pod's first container
- Uses Material "subject" icon for consistency
- Event propagation is stopped to prevent row selection
- For multi-container pods, users can still use the context menu for specific container selection

## Testing

Tested locally on:
- macOS arm64
- Windows amd64
- Single-container pods
- Multi-container pods

## UX Improvement

This feature improves developer workflow by reducing clicks needed to view pod logs - a very common operation when debugging applications in Kubernetes.

### DIFF
--- a/packages/core/src/renderer/components/workloads-pods/pods.scss
+++ b/packages/core/src/renderer/components/workloads-pods/pods.scss
@@ -57,7 +57,7 @@
     }
 
     &.logs {
-      flex-grow: 0.2;
+      flex: 0 0 29px;
     }
   }
 }


dex4er:
Hmm maybe just an icon will be even better:

image

dex4er:
Minor tweaks:

image

robertobandini (commented):
Great work also here! Very simple but useful idea! What do you think about adding also the other button to connect to a container? Maybe also in another issue.

dex4er:
I agree: we can have another PR for terminal.

PR #1303, opened 2025-10-29
Add persistent filter option for resource list views

pezzking:
## Summary

Adds an optional persistent filter feature that maintains search/filter state across view changes within the same namespace context. This significantly improves the developer workflow when debugging issues across multiple resource types.

## Problem

When debugging Kubernetes issues, developers often need to filter resources by the same criteria across different views (e.g., filtering by app name across pods, deployments, and services). Currently, each time you navigate to a different resource type, you must re-type the search filter, which is tedious and slows down the debugging workflow.

## Solution

This PR introduces a persistent filter option that:
- **Remembers your search/filter text** as you navigate between views
- **Scopes persistence per-namespace** (different namespaces have independent filter states)
- **Works in both cluster and non-cluster contexts**
- **Stores state only in memory** (session-only, not persisted to disk)
- **Can be toggled on/off** with a checkbox in the search input

### 1. New PersistentSearchStore (`persistent-search-store.injectable.ts`)
   - Manages filter state per-namespace using MobX observables
   - Integrated with the DI system
   - Session-only persistence (no disk storage)

### 2. Enhanced SearchInputUrl component
   - Added persistence toggle checkbox with "Persist" label
   - Syncs with persistent store when enabled
   - Falls back to URL param when persistence is disabled
   - Fixed namespace store injection to work in non-cluster contexts

### 3. Bug Fixes
   - Fixed dependency injection issue where \`namespaceStore\` was required in all contexts, but is only available in cluster contexts
   - Made namespace store optional and added conditional injection based on \`storesAndApisCanBeCreatedInjectionToken\`
   - Fixed reactive update logic to prevent input clearing when toggling persistence
   - Ensured smooth UX when enabling/disabling persistence
   - **Fixed checkbox styling**: Replaced native HTML checkbox with Freelens \`Checkbox\` component to match theme colors (green instead of blue)
   - **Fixed namespace switching behavior**: Added tracking to properly clear filter when switching namespaces with persistence disabled

## Use Case Example

A developer investigating a production issue:
1. Opens "Pods" view, filters by \`app=frontend\`
2. Enables persistent filter by checking the "Persist" checkbox
3. Navigates to "Deployments" view → filter automatically shows \`app=frontend\`
4. Navigates to "Services" view → filter still shows \`app=frontend\`
5. Switches namespace → filter resets (namespace-scoped)

This saves time and reduces repetitive typing during debugging sessions.

## Testing

tested on:
- ✅ macOS arm64
- ✅ Both cluster and non-cluster views
- ✅ Single and multi-namespace scenarios
- ✅ Toggle persistence on/off behavior
- ✅ Navigation between different resource types
- ✅ Namespace switching (filter isolation)
- ✅ Checkbox color matches theme (green like pod checkboxes)
- ✅ Filter clears when switching namespaces with persistence disabled

### Test Scenarios Verified
1. **Enable persistence doesn't clear input** - Typing "elastic" then clicking persist keeps the text
2. **Disable persistence clears stored value** - Clearing filter with persist off doesn't restore old values
3. **Persistence works across views** - Filter state maintains when navigating between resources
4. **Namespace isolation** - Each namespace has independent filter state
5. **Namespace switching with persistence off** - Filter properly clears when changing namespaces
6. **UI consistency** - Persist checkbox uses same green color as other checkboxes in the UI

## UI/UX

The persistent filter toggle appears as a small checkbox labeled "Persist" next to the search input. The checkbox uses the Freelens theme colors (green) to match other checkboxes in the UI. The UI is minimal and doesn't disrupt the existing layout.

---

This feature addresses a common pain point in the developer workflow and makes debugging across multiple resource types much more efficient. The implementation is clean and follows the existing architecture patterns.

### DIFF
--- a/packages/core/src/features/catalog/__snapshots__/custom-columns.test.tsx.snap
+++ b/packages/core/src/features/catalog/__snapshots__/custom-columns.test.tsx.snap
@@ -269,32 +269,55 @@ exports[`custom category columns for catalog renders 1`] = `
                     0 items
                   </div>
                   <div
-                    class="Input SearchInput focused"
+                    style="display: flex; align-items: center; gap: 8px;"
                   >
-                    <label
-                      class="input-area flex gaps align-center"
-                      id=""
-                    >
-                      <input
-                        class="input box grow"
-                        placeholder="Search..."
-                        spellcheck="false"
-                        value=""
-                      />
-                      <i
-                        class="Icon material focusable small"
-                      >
-                        <span
-                          class="icon"
-                          data-icon-name="search"
-                        >
-                          search
-                        </span>
-                      </i>
-                    </label>
                     <div
-                      class="input-info flex gaps"
-                    />
+                      class="Input SearchInput focused"
+                    >
+                      <label
+                        class="input-area flex gaps align-center"
+                        id=""
+                      >
+                        <input
+                          class="input box grow"
+                          placeholder="Search..."
+                          spellcheck="false"
+                          value=""
+                        />
+                        <i
+                          class="Icon material focusable small"
+                        >
+                          <span
+                            class="icon"
+                            data-icon-name="search"
+                          >
+                            search
+                          </span>
+                        </i>
+                      </label>
+                      <div
+                        class="input-info flex gaps"
+                      />
+                    </div>
+                    <div
+                      style="display: flex; align-items: center; white-space: nowrap;"
+                    >
+                      <label
+                        class="Checkbox flex align-center inline"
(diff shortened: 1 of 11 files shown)


dex4er:
I really like this feature.

image

I just wonder if it should be just the default option :)

pezzking:
> I really like this feature.
> 
> image
> I just wonder if it should be just the default option :)

it could easily be the default, i didn't want to intrude with my first pr to this project so i made it optional ;)

PR #1331, opened 2025-11-07
Use Turbo for faster rebuilds

### LINKED ISSUE #633: Unclear instructions for development setup

[`DEVELOPMENT.md`](https://github.com/freelensapp/freelens/blob/main/DEVELOPMENT.md) says that `pnpm dev` should spin up a development environment. As I understand it should offer a hot-reloaded Webpack server for frontend, and (optionally) spawn an Electron window pointing at that server.

However `pnpm dev` is stuck here when I run it (I have no idea if this is the correct behaviour):
```
> freelens-monorepo@ dev /home/uuu/Projects/freelens
> pnpm -r dev

Scope: 52 of 53 workspace projects
packages/kubectl-versions dev$ webpack --mode=development --watch
│ asset index.js 4.87 KiB [compared for emit] (name: index)
│ runtime modules 670 bytes 3 modules
│ cacheable modules 689 bytes
│   ./index.ts 77 bytes [built] [code generated]
│   ./build/versions.json 612 bytes [built] [code generated]
│ ./index.ts (webpack 5.99.5) compiled successfully in 432 ms
│ Type-checking in progress...
│ No typescript errors found.
└─ Running...
```

And there's no Electron window. `pnpm start` seems to run a locally-built version of Freelens (which means `pnpm build:app` must be run first).

For the record I am now at https://github.com/freelensapp/freelens/commit/d1197b891bd7e1ada2670895615e5769ae229264.

Am I missing something? Help is much appreciated.

### Why is this needed?

_No response_

MatiasRoje:
Fixes #633

**Description of changes**

- Replace the current `pnpm dev` orchestration with Turbo to speed up renderer and core watcher rebuilds and reuse cache across packages
- Added a new dev watcher to packages/core
- Dependencies:
  - Add `cross-env` and `turbo` as devDependencies to /.
  - Remove `@pmmmwh/react-refresh-webpack-plugin` and `concurrently` from freelens/
 
**Background**

- Current developer workflow: every change requires a full rebuild of the edited package (e.g., packages/core), needing as well manual parallel runs of `pnpm dev:build` and `pnpm dev:run` (in freelens/) due to the non-functional concurrent `pnpm dev`
- Reattempting the legacy Lerna+Nx approach in Freelens (and also in the legacy OpenLens) yielded poor results: caching offered minor improvements, but reloads remained slow or completely broken
- Turbo uses existing Webpack configs to trigger rebuilds in seconds, caching intermediate outputs for improved performance

**New Developer Flow**

```
# from root
pnpm dev:build   # one-time bootstrap build
Terminal A: pnpm dev    # Turbo launches webpack watchers; initial run is slower, with subsequent runs accelerated via caching
Terminal B: pnpm start-dev    # starts Electron via dev:run
```

Result: rebuilds trigger automatically on change, but the renderer still performs a full page reload and lands back on the home view after ~10–30s (on a fast machine). Turbo itself is quick and caching works well, so the experience is much better than before, even if it’s not instant HMR.

This seems like a reasonable tradeoff for now. Since the long-term plan is to upgrade to Vite and ESM, this should be a temporary measure, but it already greatly speeds up the current development workflow.


dex4er:
@MatiasRoje That's weird. Knip was telling it is not used but maybe it is some regression. I'll re-add it in a separate PR.

Issue #714, opened 2025-05-09
Restore the notice that a new version of Freelens is available


robertobandini:
### What would you like to be added?

This way all users, even those who do not directly follow Git, will at least be able to know that there is a new version available.

There should already be some code that performed this function but that needs to be re-adapted.

Issue #552, opened 2025-04-07
Autoupdate


Cenness:
### What would you like to be added?

Auto update app when new versions becomes available.  
Or at least there should be a notification about new version.

### Why is this needed?

Less congitive load on user for keeping freelens up to date.

robertobandini:
I agree, that would be awesome!

PR #1393, opened 2025-11-20
Add new version notification on Welcome Page

MatiasRoje:
It fixes: #714 and partially #552

**Description of changes**:

- Add new-version-notification injectable and wire it into the welcome page - using existing showInfoNotificationInjectable - to show a toast with a link to Github releases page when a new Freelens version is available
- Add get-latest-version request channel so the renderer can fetch the latest version, as the existing getLatestVersionInjectable is in main
- Add unit test to verify that the notification is shown only when latest > current and skipped otherwise
- Make the entire toast close-icon box clickable instead of only the icon itself

**Screenshot**:

image


robertobandini:
Thanks @MatiasRoje!!

robertobandini:
At first glance, I think it's worth keeping both notices.
The one on the home page is immediately visible and obvious to everyone, without having to go to the About page.
If you want to be picky, you could consider a "don't show again" option for the one on the home page.

PR #1400, opened 2025-11-22
Add Nedzad Winter as an adopter for Freelens

nedzad-winter:


robertobandini:
Thank you very much @dazden94!

Issue #1280, opened 2025-10-11
Custom theme and color feature


robertobandini:
### What would you like to be added?

In short, the ability to change the theme colors on Freelens and create new ones as you like.

### Why is this needed?

Many users complain that they don't have this option, especially when it comes to colors they would prefer to change.

For example:
- https://github.com/freelensapp/freelens/issues/550#issuecomment-3371177938
- https://github.com/freelensapp/freelens/issues/1237
- https://github.com/freelensapp/freelens/issues/186

robertobandini:
Hi @Excellencedev, this issue was already assigned to @gambhirsharma.
We've established that when an issue with a bounty is already assigned to someone, there's no need for someone else to submit their own PR.
I wrote this just yesterday on a dedicated page, so could be you didn't read it, I wanted to write it here too.
https://github.com/freelensapp/freelens/wiki/Fund-an-issue-or-earn-money-by-developing
so if you haven't read it, your PR could be ok, and @gambhirsharma wants to give you the bounty, that's fine.

FYI @dex4er @leo-capvano @mariomamo

gambhirsharma:
Hi @Excellencedev, I’m currently working on this issue.

Since I’m not able to work on weekdays, I’ve been contributing on weekends. The issue has been open since October, so I didn’t assume there was any urgency in raising a PR.

I also noticed that you have a few unresolved bounty PRs:

- https://github.com/tscircuit/runframe/pull/1509
- https://github.com/aqualinkorg/aqualink-app/pull/1177

I’ll review your PR as well, and I’d like to ask @robertobandini to take a look to verify its validity and overall quality. If either Roberto or I feel that your PR is solid and only needs minor adjustments, I’m happy to let you take over the issue and the bounty.

For me, the main motivation behind working on this issue isn’t the bounty, it’s that I recently started using Freelens for my job and really wanted this feature. 

Ciao!

Excellencedev:
@robertobandini what about my effort for the pr i created

Excellencedev:
@robertobandini you have not yet reviewed my pr
what's happening ?

robertobandini:
Hi @Excellencedev, when @gambhirsharma can, he'll do an initial review of your work if he wants to give you the bounty, and only then will I check.
Unless someone else on the Freelens core team wants to give their opinion on your work in the meantime.
So please wait until the review is done and don't comment further or continuously, please.
I hope I've been clear enough.

gambhirsharma:
[...]

PR #1368, opened 2025-11-15
Add Custom Theme and Color Customization Feature

Excellencedev:
This PR implements a **complete custom theme and color customization feature** for Freelens, addressing issues #1280, #1237, and #550.

#### Core Theme System Enhancements
- **Extended `LensTheme` interface** with `isCustom` and `createdAt` fields to distinguish user-created themes from built-in ones
- **Added new color properties**:
  - `colorAccent`: Explicit accent color for UI elements (#550)
  - `colorRestartedOutline`: Customizable outline color for restarted containers (#1237)

#### User Preferences Integration
- Added `customThemes` array to user preferences for storing user-created themes
- Implemented serialization/deserialization for custom themes in persistent storage
- Custom themes persist across application restarts

#### Theme Management Functions
- **`save-custom-theme.injectable.ts`**: Save or update custom themes
- **`delete-custom-theme.injectable.ts`**: Delete custom themes with automatic theme reset if active theme is deleted
- **`themes.injectable.ts`**: Dynamically merges built-in and custom themes using MobX computed values

#### User Interface Components
- **Updated Theme Selector** (`theme.tsx`):
  - Shows custom themes with "(Custom)" indicator
  - "Customize Colors..." button navigates to color editor
- **Color Customization Page** (`color-customization.tsx`):
  - Full-featured theme editor with HTML5 color pickers
  - 97+ colors organized into 5 categories (Primary, Text, Backgrounds, Status, Borders)
  - Dual input method: color picker + hex text input
  - Base theme selection (Light/Dark)
  - Theme save/load/delete functionality
  - Custom theme management interface
  - Real-time preview of color changes

#### Comprehensive Testing
- **Theme Storage Tests** (`custom-themes-storage.test.ts`):
  - Saving custom themes
  - Persistence across reloads
  - Multiple theme handling
- **Color Application Tests** (`apply-lens-theme.test.ts`):
  - CSS variable application
  - Custom color properties
  - Theme-light class toggling
- **Theme Management Tests** (`custom-theme-management.test.ts`):
  - Save/update operations
  - Delete operations
  - Active theme reset logic

#### Documentation
- **User Documentation** (`docs/custom-themes.md`):
  - Feature overview and usage guide
  - Color category descriptions
  - Troubleshooting guide
  - API documentation for extensions

Issue #550, opened 2025-04-07
Add setting for accent color


Cenness:
### What would you like to be added?

Please add an option to choose accent color. Both global and per cluster.  
If user-defined value for cluster is too hard, an option to take that accent color from generated cluster avatar/logo would be good too.
No need for color picker, just text field with hex value would be enough.

### Why is this needed?

Teal is decent, but it clashes with my os theme. And I guess easier at a glance cluster differentiation would be nice.

robertobandini:
Nice idea!

Cenness:
Turns out you can store arbitrary data in a kubeconfig.  
kubectl and freelens just ignore that field:
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: auth
    server: https://some-cluster-endpoint:6443
    accent-color: #123456       <-----------------------------
  name: some-cluster
contexts:
- context:
    cluster: some-cluster
    user: user
  name: some-cluster
current-context: some-cluster
kind: Config
preferences: {}
users:
- name: user
  user:
    client-certificate-data: cert
    client-key-data: key
```

fallenleavesgocrunch:
I'm loving Freelens but this is legit the first thing I really want to change!

robertobandini:
Personally, I thought that being able to choose colors beyond the green I chose at the beginning of Freelens was a good thing and you are confirming this.

Maybe @Shubhang can collaborate with you to this feature @mariomamo

PR #1431, opened 2025-11-29
Feat: custom color theme

gambhirsharma:
closes #1280

Selected color applies to all primary UI elements and syncs across main window and cluster iframes.

## Changes 

- **Accent color selector** with visual swatches in theme preferences
- **Reset to Default** button (shown when non-default selected)
- **Custom accent color preference** with persistence
- **Theme computation** overriding primary colors (blue, primary, buttonPrimaryBackground, menuActiveBackground, etc.)
- **IPC-based theme sync** between root frame and cluster iframes
- **Theme persistence layer** for cluster frames

## Architecture

1. **Preference Storage** - Stores/retrieves accent color from user preferences with persistence
2. **UI Components** - Color picker in Preferences panel with visual swatches
3. **Theme Computation** - Merges base theme with accent color overrides (primary, blue, buttonPrimaryBackground, etc.)
4. **Theme Application** - Applies CSS custom properties to DOM (`--primary`, `--blue`, etc.)
5. **Theme Synchronization** - IPC messaging system for cross-frame sync

**Key Design Decisions:**
- **IPC channels** for cross-origin iframe communication (cluster frames use different subdomains)
- **Multiple broadcast retries** (50ms, 150ms, 300ms) to handle async Dependency Injection container initialization
- **Global backup storage** (`window.__lastReceivedTheme`) to persist theme across document reloads and CSS variable resets

### Screenshots
image

<details>
<summary>view more screen-shot</summary>
image

image
</details>

## Testing

- [x] Color picker in Preferences 
- [x] Persists after restart 
- [x] Syncs to iframes 
- [x] Reset button 
- [x] Light/Dark themes

## Preset Colors

- **Teal** (#00a7a0) - Default
- **Green** (#4caf50)
- **Blue** (#2196f3)
- **Orange** (#ff9800)


robertobandini:
> Hi @ robertobandini, sorry for creating this PR so late 🥲

Hi @gambhirsharma no prob, on the contrary thank you very much!

> * Should the Freelens logo colour also change according to the theme it has applied?

No, I think the logo must remain as originally designed.

> * For now, [this feature](https://github.com/freelensapp/freelens/issues/550#issuecomment-2824022298) is not being implemented on this MR do we need this?

No, if the user can choose the color from the UI using your feature.

I tought you had to review [this PR](https://github.com/freelensapp/freelens/pull/1368) but @dex4er did it before you, so I suppose you opened this PR to propose your solutions.
I haven't tried it yet, but seems good, thanks!

robertobandini:
Thanks @omarluq and @gambhirsharma!
We have time as we are planning to include this feature on the 1.8.0, so we can do the poll.
@gambhirsharma if you can maybe it is better to creare the poll in the discussion here in the repo so you can share the link on Discord and everybody can vote.

robertobandini:
> We could add both a list of preset colors and a color picker, or at least a section where you can enter the color code. What do you think about adding an option in the dropdown that shows the colored square with some different colors and a "custom" label, and on click it opens the picker?

I admit that I tought the same, could be the third option on the poll, what do you think guys?

gambhirsharma:
> I admit that I tought the same, could be the third option on the poll, what do you think guys?

Yes, that's a great idea. I will implement it. 

For the poll, @robertobandini, I wasn’t sure which colors to include as options in the GitHub discussion. I posted a message on Discord to start the conversation, and once we see which colors are most popular, we can add them to the presets.

Also, @robertobandini, could you review this MR or assign someone to take a look at it

robertobandini:
@mariomamo If you would like you can review this too, just let know, thanks!

packages/core/src/features/pod-logs/__snapshots__/download-logs.test.tsx.snap
mariomamo: Why this snap has been deleted?

gambhirsharma: Sorry, I wasn’t able to update the snapshots after the new UI changes. So, in my confusion, I thought deleting the existing snapshots and regenerating them from the updated UI might be the right approach.

Could you please explain how snapshot testing works and what the correct way is to update snapshots when the UI changes?

```bash
# Delete all snapshots first
find . -type d -name __snapshots__ -exec rm -rf {} + 2>/dev/null

# Then run tests with update flag
pnpm test:unit:updatesnapshot
```

omarluq: hey @gambhirsharma to update a snap
- run `cd packages/{package-where-snap-is}`
- run `CI=false pnpm test:unit -- -u` to run all tests and update snaps or alternatively you can run `pnpm jest <testfile> -u`
- after that just stage and commit the updated snap file

lmk if you face any issues with this 👍🏼

packages/core/src/features/welcome/__snapshots__/navigation-using-application-menu.test.ts.snap
```diff
@@ -549,81 +549,171 @@ exports[`welcome - navigation using application menu when navigated somewhere el
...
+                                class="css-tj5bde-Svg"
+                                focusable="false"
+                                height="20"
+                                viewBox="0 0 20 20"
+                                width="20"
+                              >
+                                <path
+                                  d="M4.516 7.548c0.436-0.446 1.043-0.481 1.576 0l3.908 3.747 3.908-3.747c0.533-0.481 1.141-0.446 1.574 0 0.436 0.445 0.408 1.197 0 1.615-0.406 0.418-4.695 4.502-4.695 4.502-0.217 0.223-0.502 0.335-0.787 0.335s-0.57-0.112-0.789-0.335c0 0-4.287-4.084-4.695-4.502s-0.436-1.17 0-1.615z"
+                                />
+                              </svg>
```
omarluq: can we move this to `packages/ui-components/icon/assets/{name-svg}.svg‎` instead on inline plz

gambhirsharma:
> @gambhirsharma Is the DI syncing process not working out of the box for this feature? Did you find some problem? An alternative to the current implementation would be to have userPreferences state to emit some "themeChanged" event and let the clusterFrames and UI elements react to it by themselves.G

hi, @MatiasRoje, DI syncing wasn't working for me. I got stuck on it, so I decided to implement this multi-tries & global backup storage. Initially, I only wanted the feature to work & didn't think much about anti-pattern.

I will try to look into the project again and adopt a better approach. 

Sorry for the delays just completed the exams today, so I will try to complete this PR ASAP

MatiasRoje:
Hey @gambhirsharma, good progress so far. We want to get this feature merged soon, so let's keep working on it.

A few blocking points to address before we merge:

- Please remove the separate theme IPC/update path (`themeChangedChannel` and related listener/wiring). User preferences storage already syncs state across frames, so this creates a duplicate propagation path for the same data and is not consistent with the rest of the codebase.
- Make accent overrides explicit by key while iterating through the baseTheme.color objects. Matching the theme color override by raw hex value (`value === #00a7a0`) is a bit fragile and could have side effects for future themes or extensions. It would be better to define a fixed list beforehand and use that for the matching: 
const accentKeys: LensColorName[] = [
    "primary",
    "menuActiveBackground",
    "buttonPrimaryBackground",
    // ...
  ];
- Add a couple of reliable tests for this feature.

A note on the accent colors:

- Blue looks fine. Green and orange are already being used as the colorOk and colorWarning in the light/dark themes. I do like them as options but we should go for different shades or tints that are unique and appropriate for a primary color.
Please consider good readability over white and dark backgrounds.

The feature is close and on the right track. Once these points are addressed I think we are ready to merge. I would add a couple of more nice-looking color options and don't introduce the color-picker in this PR. I would focus on what we have already and improve it over time if the community requests it. Would you agree with that @robertobandini ?

gambhirsharma:
Thanks @MatiasRoje, for the review! I will push the changes ASAP

> The feature is close and on the right track. Once these points are addressed I think we are ready to merge. I would add a couple of more nice-looking color options and don't introduce the color-picker in this PR. I would focus on what we have already and improve it over time if the community requests it. Would you agree with that @robertobandini ?

I believe this would be a better approach to go with for now. I have implemented this color picker in my local branch, but I don't feel confident with the implementation; that's the reason why I didn't add the commits in this PR. Let me know what you think @robertobandini, about it.

robertobandini:
Hi guys @gambhirsharma @MatiasRoje, thanks!
I agree with you; it's best to finalize it now, as Matias also said, and then further improve it based on community requests.

robertobandini:
Hi @gambhirsharma! Thanks about keep working on this until we reach the goal!

gambhirsharma:
@robertobandini @dex4er 
Hey! Could you please review it!

I belive its ready now. Thanks

MatiasRoje:
Hey @gambhirsharma , thanks! The separate theme IPC path is gone and using the user preferences storage is correct. The active theme tests are also a good addition. Colors also look fine, some of them better on light than on dark or the other way around, but overall it is fine.

I think that the accent override approach is acceptable: replacing only the default teal on a known set of keys preserves the existing differences between the light and dark themes. What I would change is to make that intent explicit. Right now it reads a bit like “these keys are always controlled by the accent color”, but the actual behavior is “only replace the default teal when it appears on these keys”. Could we rename/comment this accordingly and add coverage against the real built-in light/dark themes? For example, the dark theme should keep sidebarSubmenuActiveColor as white, while the light theme changes it because it currently uses teal there. That would make the behavior clear and protect it from accidental changes later.

And a really small issue I would also fix while we are here: please avoid storing #00a7a0 when the user selects the default Teal option. Selecting the default should clear customAccentColor, following the existing pattern of not persisting default preference values in the storage.

Overall it looks good to me. Let's wait for feedback from the others.

robertobandini:
Hi @gambhirsharma, thanks again for all the work done here, the PR is really close.

As I wrote on #1280, we plan to finalize this feature soon. Let me share the
plan, because the timing is now linked to the Freelens v2 release.

We are finalizing v2 in these weeks (current estimate is about 2-3 weeks,
@dex4er updates it as the work goes on, see #2279). Since v2 changes the
foundations of the codebase (ESM and more), it makes no sense to complete this
PR twice: first on the current base and then again after a big rebase. So the
plan is to complete it once, directly on top of v2.

Concretely, two steps:

1. Could you confirm us by [DATA, es. August 3] that you are still available
   to finish this PR after the v2 release?
2. Once v2 is out, you would have [es. two weeks] to rebase the branch and
   complete the remaining points from @MatiasRoje's review (remove the extra
   IPC path, explicit accent keys list, tests). We can run Claude on your
   branch to help with the rebase and the remaining points, keeping your
   authorship: just ask.

In the meantime, if you want, you can start looking at the new v2 setup, so
you arrive ready.

If we don't hear from you, or the PR is not completed in that window, the core
team will finalize it with Claude's help. In any case, the bounty stays yours:
you will collect it anyway, as a recognition of the contribution you have
given so far.

This PR stays focused on the accent color feature. For the full custom themes
system, see #1847: I'll leave a comment there about how we see the next steps.

I hope I've been clear enough, and thanks again for your time on this!

FYI @dex4er @MatiasRoje @omarluq @mariomamo

gambhirsharma:
Thanks for the update, @robertobandini.

Yes, I can confirm that I'll be available to finish this PR after the v2 release. It seems like a logical decision.

Until then, I'll keep an eye on the Discord and GitHub discussions related to the release and the arch changes.

PR #1538, opened 2026-01-12
auto suspend

regal-fang:
This code solves the problem that when using sso, the computer locks the screen and pops up the web page to login


dex4er:
@regal-fang Actually I don't understand. Does it suspend re-authentication to Kubernetes if the suspend mode on the OS is detected or what?

Could you explain a little more what this PR fixes exactly? Thanks!

regal-fang:
> @regal-fang Actually I don't understand. Does it suspend re-authentication to Kubernetes if the suspend mode on the OS is detected or what?
> 
> Could you explain a little more what this PR fixes exactly? Thanks!

Hi @dex4er 
My use case is as follows: I use freelens to connect to EKS via SSO. When freelens is running, if I simply close my MacBook or click lock/sleep, freelens continues to periodically attempt connecting to EKS in the background. This prevents the MacBook from entering sleep mode. Moreover, once the SSO session expires, login windows keep popping up repeatedly (about every 15 minutes)—dozens of such windows can appear overnight. The change I made is intended to fix this issue.

In addition, i want to be able to respond to the computer's sleep state at the same time, otherwise the background operation will prevent the battery power from sleeping

I am currently using this modified version, and it has been running well for several months now.

Thanks

dex4er:
@claude review the MR especially for regressions

robertobandini:
Hi @regal-fang, thanks for the PR and for explaining the EKS/SSO use case. It looks like other users are seeing a similar issue in #1571, so I don't want this contribution to get lost.

@dex4er, did Claude's review help identify a good way forward here? @regal-fang, was the feedback clear and useful enough to update the PR, or is there anything we should clarify?

It would be good to agree on the next steps between this suspend approach and the backoff proposed in #1571.

dex4er:
@robertobandini We have some exponential backoff implemented as some kind of mitigation for this problem. Currently, I don't have any EKS cluster to test how it really works, so no way of real checking if the problem still occurs. The implementation has some real problems, as Claude found them.

I would leave this solution right now as the main rule is: the fix is good if it doesn't break more things. Here: not so sure :)

Issue #1539, opened 2026-01-12
[FeatureRequest] Warning UI Element if pod is not ready and has warning events


robertdyr:
### What would you like to be added?

Warning marker for pods that are not ready and have warning events. Lens has this feature: 

<img width="523" height="164" alt="Image" src="https://github.com/user-attachments/assets/70733282-98a1-44db-88ef-078eb7d30e22" />

<img width="518" height="216" alt="Image" src="https://github.com/user-attachments/assets/736990cb-ac07-4493-ab7a-ea59d6246101" />

<img width="1661" height="175" alt="Image" src="https://github.com/user-attachments/assets/05f87994-462b-4c29-8908-01fa5c86e1f0" />

### Why is this needed?

I realized this is a feature i heavily rely upon. I tried switching from lens to freelens and this was one of the features i really missed. It allows me to open the cluster and immediatly see where things are going wrong. Simply the containers state is not enough, because we could just have slow starting pods/containers. The visual cue of a) container is not ready and b) there is a warning event helps debugging a cluster quickly. This UI component is really helpful in quickly identifying problematic pods.

I imagine to implement this we need to watch the conditions: 

```yaml
status:
  phase: Running
  conditions:
    - type: PodReadyToStartContainers
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-01-12T14:15:09Z'
    - type: Initialized
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-01-12T14:15:07Z'
    - type: Ready
      status: 'False'
      lastProbeTime: null
      lastTransitionTime: '2026-01-12T14:15:07Z'
      reason: ContainersNotReady
      message: 'containers with unready status: [app]'
    - type: ContainersReady
      status: 'False'
      lastProbeTime: null
      lastTransitionTime: '2026-01-12T14:15:07Z'
      reason: ContainersNotReady
      message: 'containers with unready status: [app]'
    - type: PodScheduled
      status: 'True'
      lastProbeTime: null
      lastTransitionTime: '2026-01-12T14:15:07Z'
```
and then find the last event of type `Warning` in the events of the pod.

<img width="1661" height="175" alt="Image" src="https://github.com/user-attachments/assets/6abcbb04-6e44-4175-b208-0932be37c304" />

Not sure which conditions exactly are needed to watch in case of initContainers and multi container pods.

PR #1540, opened 2026-01-13
Add warning column for pods with K8s warning events

robertdyr:
Implements #1539

Screenshot 2026-01-13 114259
Screenshot 2026-01-13 114140

If someone wants to see the behaviour in UI here are four cases: 

[never-ready.yaml](https://github.com/user-attachments/files/24587310/never-ready.yaml)
[delayed-ready.yaml](https://github.com/user-attachments/files/24587277/delayed-ready.yaml)
[live-and-ready-alternate-failing-then-succeed.yaml](https://github.com/user-attachments/files/24587274/live-and-ready-alternate-failing-then-succeed.yaml)
[multi-container.yaml](https://github.com/user-attachments/files/24587275/multi-container.yaml)

I realized that Lens also has this feature for deployments and deamonsets, which may also be useful. However i think i will leave this as is for now and maybe add those features in a seperate PR.


MatiasRoje (approved):

MatiasRoje:
LGTM

PR #1543, opened 2026-01-16
New preference to toggle delete protection and check for update

mervynzhang:
New preference to toggle delete protection and check for update
image


MatiasRoje:
@mervynhang Thank you very much for your contribution.

Check for new version on startup is working as expected and there are also a new test for the functionality. Great!

Unfortunately, you can still delete resources in a few places even when "Allow deleting Kubernetes resources" is toggled off. To fully honor the new delete protection, we would need to apply the same guard in some other spots:

1) Bulk delete button (minus icon that appears when you click on the checkbox left to the resource name)
   - packages/core/src/renderer/components/item-object-list/content.tsx: (AddRemoveButtons onRemove). This covers all list views including Namespaces and Helm Releases.
 
2) Namespaces has it seems a custom delete flow
   - packages/core/src/renderer/components/namespaces/route.tsx (removeItems, removeSelectedItems)

3) Helm Releases has also a custom delete flow
   - packages/core/src/renderer/components/helm-releases/releases.tsx (removeSelectedItems)

Already gated:
- Single-item context menu for standard Kube objects. 

It would be good to centralize the allowDelete logic and add tests for this feature. Thank you!

robertobandini:
Hi @mervynzhang thanks about your contribution!
The option to check for updates seems useful.
I'm also wondering, since I don't have time to check, what is the effect the delete protection option?
In these cases, please explain the use case and provide screenshots so we can immediately understand the purpose.
It could be useful if, for example, it asks "Are you sure?" when you delete something.

@MatiasRoje @omarluq 
If we're unsure about its usefulness, we can have the community vote on it.

packages/core/src/features/preferences/renderer/preference-items/application/check-for-updates/check-for-updates.tsx
```diff
@@ -0,0 +1,33 @@
...
+import userPreferencesStateInjectable from "../../../../../user-preferences/common/state.injectable";
+
+import type { UserPreferencesState } from "../../../../../user-preferences/common/state.injectable";
+
+interface Dependencies {
+  state: UserPreferencesState;
+}
+
+const NonInjectedCheckForUpdates = observer(({ state }: Dependencies) => (
+  <section id="other">
```
omarluq: Same duplicate `id="other"` here, can we use
```suggestion
  <section id="check-for-updates">
```

packages/core/src/renderer/components/item-object-list/__tests__/delete-protection.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+        allowDelete: true,
+      }));
+    });
+
+    it("shows the remove button when items are selected", () => {
+      const props = createProps();
+
+      render(<ItemListLayoutContent {...props} />);
+
+      const removeButton = document.querySelector(".remove-button");
```
omarluq: This test uses `.remove-button` (CSS class). 
Can we query by a stable id/testid instead (e.g. `data-testid="remove-button"` or `role`+`label`) to avoid brittle selectors?

packages/core/src/features/preferences/renderer/preference-items/application/delete-protection/delete-protection.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface Dependencies {
+  state: UserPreferencesState;
+}
+
+const NonInjectedDeleteProtection = observer(({ state }: Dependencies) => (
+  <section id="other">
+    <SubTitle title="Delete Protection" />
+    <Switch checked={state.allowDelete ?? true} onChange={() => (state.allowDelete = !state.allowDelete)}>
+      Allow deleting Kubernetes resources
+    </Switch>
```
omarluq: Small bug: when allowDelete is undefined (default), `onChange={() => (state.allowDelete = !state.allowDelete)}` evaluates to `true`, so the first click is a no‑op.


```suggestion
    const value = state.allowDelete ?? true;
    <Switch
      checked={value}
      onChange={() => (state.allowDelete = !value)}
    >
      Allow deleting Kubernetes resources
    </Switch>
```

omarluq (changes requested):
Hey @mervynhang great work! thanks for contributing, I just have a few comments for your consideration, thanks.

robertobandini:
> Hi @robertobandini if we enable "Allow deleting Kubernetes resource", the Delete button will show in the resource 

This seems a bit extreme to me. Would it be possible to ask the user if they are sure they want to delete something before doing so when clicking the delete buttons if protection is enabled?

PR #1557, opened 2026-01-27
Respect default container

davidmayr:
### Changes
Kubectl allows specifying a default container for logs, exec etc. using the annotation "kubectl.kubernetes.io/default-container"

This PR adds support for this feature in freelens.

### Motivation

We deploy Agones in our Kubernetes cluster and it automatically attaches a sidecar. Clicking the logs button always opens the sidecar container, as it's the container with index 0. However, you'd likely never need to see the Agones sidecar output, so this has been a very big annoyance on our dev cluster. 

When we found out that the default container annotation exists, we immediately deployed it but unfortunately found out that there is no support for this in Freelens.


dex4er:
I agree, the feature is not complete yet

davidmayr:
I completely forgot about this as I wasn't doing that much in Kubernetes recently. But I'll try to update this in the next few weeks

robertobandini:
Thanks @davidmayr!

davidmayr:
@MatiasRoje Where should I put the default-container-helper.ts? "dock/logs" might not be the best choice since pod-menu-item also exists for other things and not only logs, so I think I should choose a better location. 

But from a quick glance, the only idea I'd have that would be general enough would be renderer/utils maybe?

MatiasRoje:
Hey @davidmayr ! Good observation. I would not put it in the generic renderer/utils though: this is more of a K8s pod container behavior, not a generic renderer utility for multiple resources. My suggestion would be to move it into the Pod domain somewhere, maybe as a helper/method in packages/kube-object/src/specifics/pod.ts (e.g. a getDefaultContainer() method there) or otherwise under packages/core/src/renderer/components/workloads-pods/.
Let me know when the PR is ready to review. Thank you.

dex4er:
@claude Fix this PR in the new PR: we expect that clicking on the "Attach to pod", "Pod shell" and "Pod logs" icon (not a menu item) will do an action on the container that is marked with `kubectl.kubernetes.io/default-container` annotation. If the annotation points to the non-existing container or is missing then it should restore the default behavior (the first container).

PR #1571, opened 2026-02-05
fix: add exponential backoff for auth failures to prevent browser tab…

mitchross:
When using OIDC auth plugins like kubelogin, expired tokens cause the 30-second refresh timer to repeatedly trigger exec-based credential plugins, opening a new browser tab each time. This results in dozens of tabs accumulating when a user is away.

Add auth failure tracking with exponential backoff (1min, 5min, then stop after 3 consecutive failures) to ClusterConnection. Only auth- related errors (4xx status, credential fetch failures) trigger backoff; network timeouts and server errors do not. Manual reconnect via the UI resets the backoff state. Also add a concurrent refresh guard to prevent overlapping API calls.

Includes 13 unit tests covering backoff escalation, max retry stopping, recovery reset, non-auth error passthrough, and concurrent refresh prevention.

https://claude.ai/code/session_01Wt5xuSzeti8pWERNCDqPdQ


dex4er:
I think #1538 tried to solve this problem in another way, but this exponential backoff logic looks like a correct enhancement that we could get it as first.

dex4er:
@claude Check if the implementation is correct and without regression. Also check if interacts with #1538 that was merged earlier.

robertobandini:
Hi @mitchross! Sorry this one went quiet for a while. The problem you reported is real and we still want to solve it, those piling browser tabs are a terrible experience!

Quick recap of where we stand:

- Claude's review above found a few concrete issues. The main one is that some auth error patterns are too broad: `AbortError` and `context deadline exceeded` can come from plain network hiccups, and backing off on those would suppress retries exactly when we want them. The other two points are enforcing the backoff inside `refresh()` itself and adding test coverage for the `isAuthRelatedError` path.
- Since then `main` switched from Jest to Vitest (#2113), so the new test file would also need porting.

Would you like to update the PR along these lines? No pressure on timing! If you'd rather not, just say so and we can take it over from here, you already kindly offered the "close or redo" option. Either way, thanks a lot for reporting this with such a clear repro 🙏

robertobandini:
Hi @mitchross! Thanks again for this one, the problem was real and your backoff is the base of the fix.

While verifying it on the v2 main branch we found that the error patterns could never match there: the connection check only sees "Internal Server Error" from the proxy, or the abort reason of the timeout, so the plugin was still invoked every 30 seconds. I opened #2459 which keeps your three commits as they are and adds on top of them a change to keep the reason of the proxy in the request errors, a precise detection (credential plugin failure reported by the proxy, or a timeout on a cluster whose kubeconfig user runs an exec plugin) and the review points, with tests and an end to end check with a fake exec plugin. You are credited as co-author.

Closing this one in favour of #2459, and thanks for the "close or redo" offer, it made this easy.

mitchross:
No problem happy to help!

PR #1646, opened 2026-02-15
Fix persistent search dropping characters while typing

pezzking:
## Summary

- Fixes a race condition in the persistent search input where characters would be dropped while typing
- The MobX reaction that syncs `inputVal` from the persisted store would fire when the URL param debounce (250ms) completed before the storage write debounce (500ms), overwriting `inputVal` with the stale persisted value
- Adds a `userTyping` flag to prevent the reaction from overwriting `inputVal` while debounced writes are in flight
- Aligns both debounce timers to 250ms so storage stays in sync with the URL param

## Test plan

- [ ] Enable "Link search (shared)" in any resource list view (e.g. Pods)
- [ ] Type quickly in the search input — characters should no longer be dropped
- [ ] Verify search still syncs correctly when switching between namespaces
- [ ] Verify clearing the search input works as expected
- [ ] Verify toggling link/unlink preserves the search value


robertobandini:
Thanks @pezzking and @MatiasRoje!

PR #1647, opened 2026-02-15
Add OpenShift Prometheus support with bearer token authentication

pezzking:
## Summary

- Adds a new **OpenShift** Prometheus provider that auto-detects `prometheus-k8s` in the `openshift-monitoring` namespace
- Adds optional `directUrl` and `bearerToken` fields to Prometheus preferences, enabling direct requests that bypass the Kubernetes API service proxy
- Adds OpenShift-specific UI settings with contextual help text for configuring the Prometheus route URL and service account bearer token

## Problem

On OpenShift/OKD clusters, Prometheus runs behind a `kube-rbac-proxy` sidecar that requires bearer token authentication. The Kubernetes API service proxy path (`/api/v1/namespaces/.../services/.../proxy/...`) cannot forward `Authorization` headers to backend services, making it impossible to query Prometheus metrics through Freelens.

## Solution

When an OpenShift Prometheus `directUrl` is configured (typically an OpenShift Route), `get-metrics.injectable.ts` makes a direct HTTPS request to that URL with the bearer token in the `Authorization` header, completely bypassing the K8s API service proxy.

### Usage

1. Select **OpenShift** as the Prometheus provider in cluster settings
2. Enter the Prometheus route URL: `oc get route prometheus-k8s -n openshift-monitoring -o jsonpath='https://{.spec.host}'`
3. Generate a long-lived service account token: `oc create token <service-account> -n openshift-monitoring --duration=8760h`
4. Metrics will appear in cluster, node, and pod dashboards

### Files changed

| File | Change |
|------|--------|
| `packages/core/src/common/cluster-types.ts` | Add `directUrl` and `bearerToken` to `ClusterPrometheusPreferences` |
| `packages/core/src/main/get-metrics.injectable.ts` | Add direct fetch path with bearer token when `directUrl` is set |
| `packages/core/src/renderer/components/cluster-settings/prometheus-setting.tsx` | Add OpenShift-specific UI fields with contextual hints |
| `packages/technical-features/prometheus/src/openshift-provider.injectable.ts` | New OpenShift provider (reuses Operator query format) |
| `packages/technical-features/prometheus/src/register-injectables.ts` | Register the OpenShift provider |

## Test plan

- [ ] Select OpenShift provider on an OpenShift/OKD cluster
- [ ] Configure Prometheus route URL and bearer token
- [ ] Verify settings persist across page navigation
- [ ] Verify metrics load in cluster overview, node, and pod dashboards
- [ ] Verify non-OpenShift providers still work unchanged (no regression)
- [ ] Verify Auto Detect still works on non-OpenShift clusters

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -58,7 +58,7 @@ lint:
       files:
         - sass
   enabled:
-    - actionlint@1.7.10
+    - actionlint@1.7.11
     - biome@2.3.14! # datasource=npm depName=@biomejs/biome
     - git-diff-check
     - markdownlint@0.47.0
--- a/freelens/electron-builder.yml
+++ b/freelens/electron-builder.yml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/electron-userland/electron-builder/refs/heads/master/packages/app-builder-lib/scheme.json
 
-electronVersion: 39.5.2 # datasource=npm depName=electron
+electronVersion: 39.6.1 # datasource=npm depName=electron
 
 ## Run `pnpm electron-rebuild` instead
 npmRebuild: false
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -1,40 +1,42 @@
 {
   "name": "freelens",
-  "private": true,
-  "productName": "Freelens",
-  "description": "Freelens - Free IDE for Kubernetes",
-  "homepage": "https://freelens.app",
   "version": "1.8.2-0",
+  "private": true,
+  "description": "Freelens - Free IDE for Kubernetes",
+  "keywords": [],
+  "homepage": "https://freelens.app",
+  "bugs": {
+    "url": "https://github.com/freelensapp/freelens/issues"
+  },
   "repository": {
     "type": "git",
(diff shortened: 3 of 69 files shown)


packages/core/src/renderer/components/cluster-settings/prometheus-setting.tsx
```diff
@@ -119,7 +132,26 @@ class NonInjectedClusterPrometheusSetting extends React.Component<ClusterPrometh
...
+        service: parsed[1] || "",
+        port: parseInt(parsed[2]) || 0,
+        prefix: this.sanitizePrefix(this.customPrefix),
+        https: this.useHttps,
+        directUrl: this.directUrl,
+        bearerToken: this.bearerToken || undefined,
+      };
+    }
+
+    if (!this.path) {
```
omarluq: it seems that we are interdocung a bit of duplication here can we tighten up the code plz perhaps extract a a small factory helper for the return value, also perhaps we should create a concrete type for the return obj

omarluq (approved):
tysm! this lgtm! 

cc: @dex4er @MatiasRoje @robertobandini  second opinion before merging

PR #1651, opened 2026-02-16
Develop

me0wster:


mariomamo:
Hi @me0wster, thank you for your effort!

It looks like you're introducing an AI chat feature in Freelens, which is already available as an [official plugin](https://github.com/freelensapp/freelens-ai-extension).

We chose to keep it as an extension so users can decide whether they want to enable AI capabilities or not.

If you’d like to improve it, we’d really appreciate contributions directly to the freelens-ai-extension repository.

Thank you for your understanding. For future issues, including a short description would be greatly appreciated.

robertobandini:
@me0wster Thank anyway you about your interest in Freelens, you are welcome to contribute.

Issue #1170, opened 2025-09-11
Log selection reset by scrolling


WalterWolf49:
### Describe the bug

The text selection in the log view/window gets reset by scrolling (either by automatic scrolling with the text selection or by manually scrolling with a mouse wheel).
This means only the currently visible logs can be selected and copied, anything outside the viewport cannot be selected and copied.

### Reproduction steps

Open e.g. a Pod Log and select some text at any point in the log and scroll. Scrolling will remove the selection

### Expected behavior

Text can be selected and stays selected when scrolling, allowing to select more lines from a log than currently visible / currently in the viewport

### Screenshots

If really required, I can try and provide a screenshot (through dummy pod logs or similar, but that may take some time). All current logs I have access to may contain sensitive content, hence no screenshot. 

### Freelens version

1.5.3

### Operating system type and version

MacOS 15.6.1

### Installation method

homebrew

### Kubernetes cluster version used with Freelens

1.34

PR #1674, opened 2026-02-26
Add word wrap toggle to pod logs

### LINKED ISSUE #1563: Add “Word Wrap” toggle for Pod Logs view

**Description**

While viewing pod logs in Freelens, long log lines are currently rendered as a single horizontal line. This forces users to scroll horizontally to read the full log entry, which makes log analysis difficult—especially for JSON logs, stack traces, or verbose application output.

It would be very helpful to have a Word Wrap toggle in the logs panel so logs can be wrapped within the visible viewport.

**Current Behavior**
	•	Log lines are displayed in a single horizontal line.
	•	Users must scroll horizontally to view long log entries.
	•	This reduces readability and usability during debugging.

**Expected Behavior**
	•	Add a “Word Wrap” checkbox/toggle at the bottom of the logs view (similar to other log viewers).
	•	When enabled:
	•	Log lines should wrap within the visible width.
	•	No horizontal scrolling should be required.
	•	When disabled:
	•	Keep the current single-line behavior.

Proposed UI Location: Bottom of the logs panel, alongside other controls (see highlighted area in screenshot).

Image

### Why is this needed?

Improves log readability
Makes JSON and stack traces much easier to analyze
Aligns with UX patterns used in other Kubernetes tools and IDEs

MatiasRoje:
It fixes #1563

## Description

This PR adds a Word wrap toggle to pod logs so long lines are easier to read without horizontal overflow.

When word wrap is enabled, logs now use dynamic row heights so wrapped lines are fully visible in the virtualized list.  
In simple terms: we estimate row height from visible text length and container width, and then adjust if the rendered row needs extra space to avoid overlapping (it was the case sometimes, specially with long log lines).

I also improved the "go to bottom" behavior and button style, so jumping to bottom reliably shows the latest log line in view (or very close to it, eventually a new line might be generated while scrolling).

## What changed

- Added Word wrap control in logs UI.
- Enabled wrapped log rendering with CSS class.
- Implemented dynamic row-height calculation for wrapped logs.
- Improved row-height correction after render (for edge cases).
- Updated bottom scroll behavior to align to the end of the last line.
- Added/updated tests for:
  - Word wrap toggle behavior in controls
  - Word-wrap class application in log rows

## Notes

- Word wrap is true by default. No persistence was added in this PR for log controls, as the existing options don't have persistence at the moment and adding it only for word wrap would be inconsistent. Show timestamp should also be persistent in that case, not so sure about Show previous terminated container.
If other maintainers prefer persistence across pod logs/sessions for those options, I can gladly add that in a follow-up PR.

- My intention is to migrate logs from the current virtual-list solution to xterm in the long run, as I really dislike the current copy/paste experience (as mentioned in #1170), and that problem is not solvable with the current virtual-list architecture. I also want to improve logs UX overall (some UI refinements, richer search options) and extend logs support to workloads like deployments/replicasets/statefulsets, plus other open issues reported by users.

## Screenshot

image

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -58,7 +58,7 @@ lint:
       files:
         - sass
   enabled:
-    - actionlint@1.7.10
+    - actionlint@1.7.11
     - biome@2.3.14! # datasource=npm depName=@biomejs/biome
     - git-diff-check
     - markdownlint@0.47.0
--- a/freelens/electron-builder.yml
+++ b/freelens/electron-builder.yml
@@ -1,6 +1,6 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/electron-userland/electron-builder/refs/heads/master/packages/app-builder-lib/scheme.json
 
-electronVersion: 39.5.2 # datasource=npm depName=electron
+electronVersion: 39.6.1 # datasource=npm depName=electron
 
 ## Run `pnpm electron-rebuild` instead
 npmRebuild: false
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -1,40 +1,42 @@
 {
   "name": "freelens",
-  "private": true,
-  "productName": "Freelens",
-  "description": "Freelens - Free IDE for Kubernetes",
-  "homepage": "https://freelens.app",
   "version": "1.8.2-0",
+  "private": true,
+  "description": "Freelens - Free IDE for Kubernetes",
+  "keywords": [],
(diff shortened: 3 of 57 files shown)


MatiasRoje:
@dex4er @omarluq @mariomamo @robertobandini 

It would be great to get a second opinion here before merging.

MatiasRoje:
> @MatiasRoje out of curiosity asking, which release we going to get this feature

In the next release, v1.9.0.

PR #1685, opened 2026-03-07
Fix namespace filter pasted selection when "All Namespaces" is active

### LINKED ISSUE #1677: "Copy & Paste" in "namespace filter" not working
### Describe the bug

Kind of the [same](https://github.com/freelensapp/freelens/issues/721) but in "namespace filter".
When you paste a namespace name inside, you are not able to select it.

Image

### Reproduction steps

1. Go to `Workloads -> Pods`

Image

2. Take `kube-system` into your clipboard (`CTRL+C`)
3. Paste your clipboard (`CTRL+V`) content into the "namespace filter field" and try to select the value

Image

4. I would expect that `kube-system` is selected as the only namespace

Image

### Expected behavior

I would expect that `kube-system` is selected as the only namespace

### Screenshots

_No response_

### Freelens version

1.8.1

_No response_

### Installation method

_No response_

_No response_

### Logs

_No response_

### Kubeconfig

_No response_

### Additional context

_No response_

Tal-Naeh:
## Summary

Fixes #1677

When a user pastes (or types) a namespace name that exactly matches an existing namespace into the namespace filter input, it is now automatically selected and the menu closes.

Previously, pasting a namespace name into the filter had no effect — the user could see the filtered option in the dropdown but was unable to select it via paste.

### Changes

- Added `onInputChange` handler to `NamespaceSelectFilterModel` that performs a case-insensitive exact match against available namespaces and auto-selects on match
- Wired `onInputChange` into the `Select` component in `NamespaceSelectFilter`
- Added unit tests covering both the exact-match (auto-select) and partial-match (no auto-select) cases

## Test plan

- [x] Open Workloads → Pods
- [x] Copy a namespace name (e.g. `kube-system`) to clipboard
- [x] Click the namespace filter and paste — the namespace should be auto-selected
- [x] Verify partial text (e.g. `kube-`) keeps the menu open without auto-selecting
- [x] Run unit tests: `pnpm test:unit --filter=@freelensapp/core -- --testPathPattern=namespace-select-filter`


MatiasRoje:
Hm. The namespace filter is multi-select and auto-selecting makes text input change the selection state before explicit user confirmation. I think that can conflict with expected multi-select behavior and it's a bit too much.

I’d suggest this behavior instead:

1. Typing/pasting should only filter options.
2. Selection should happen only on explicit accept (Enter or clicking an option).
3. Acceptance behavior should depend on current selection state:
      - If all namespaces are currently selected, accepting one namespace after copy-paste should switch to only that namespace. Currently that's not the case, I'm seeing my copy-pasted namespace being removed from the selection, which is weird. This is what #1677 is highlighting and I agree it should be fixed. If I type the namespace name myself and select it, it works as expected and selects the namespace instead of removing it. 
      - If some namespaces are already selected (but no all), accepting a namespace after copy-paste should add it to the current selection if holding ctrl/meta+enter (or ctrl/meta+clicking the option) or make it the only selected namespace if only enter (or clicking). The behavior here is also inconsistent on type, same as with copy-paste.
      
Incidentally, the filter stop working sometimes for some reason if you already use it once and the modal closes (e.g. it happens on multi-select). The filter keeps focused but you don't see results anymore if you type. I might open a second issue about that.

MatiasRoje:
The selection modifier key is the same as the paste key, that's the core of the issue.
For clarity, I made a small naming refactor to use the pressed key as the state driver (isMultiSelection -> isSelectionModifierPressed) to better match what's actually happening and added a focused test in the non `when multi-selection key is pressed` path.
Otherwise, the fix was already correct. Thanks @Tal-Naeh !

PR #1693, opened 2026-03-16
Add custom theme and color configuration feature

addidea63:
Closes #1280

### Summary

The feature introduces a custom theme and color configuration to improve the user experience by providing theme flexibility. It allows users to select different themes, which are handled through a new theme configuration system. The solution is broken into multiple components to maintain modularity and improve maintainability.

### Detailed breakdown

1. **themeConfig.js**: This module is responsible for holding the theme configurations. It defines the structure of the themes and provides an API to retrieve and update the active theme.
   - The module exposes a `getTheme` function to fetch the current theme, and an `updateTheme` function for applying a new theme. These functions are used by the `ThemeProvider` component.
   - Configurations for various themes (like dark and light mode) are defined here, making it easy to extend in the future.

2. **themeConfig.test.js**: A corresponding test suite has been created to ensure the theme configuration works as expected. It includes unit tests for both theme retrieval and updating, ensuring that the theme state is correctly modified and can be fetched at any point.
   - The tests simulate user interactions with theme switches to validate that changes persist as expected.

3. **ThemeProvider.js**: This component manages the theme state across the application. It acts as a context provider, allowing child components to access the active theme. The state is managed using React’s `useState` and `useEffect` hooks for seamless updates.
   - The `ThemeProvider` ensures that the theme is applied globally, allowing all parts of the app to react to changes in the theme.
   - It also integrates with local storage to persist the selected theme between page reloads, improving the user experience.

4. **ThemeSwitcher.js**: This component allows users to switch between available themes. It is designed to be easily added to any part of the app, offering a toggle or dropdown menu to switch themes.
   - The `ThemeSwitcher` listens for user actions and triggers the `updateTheme` function from the `themeConfig.js` module.
   - The switcher also ensures that the selected theme is persisted in local storage.

### Technical considerations

- **Component structure**: I opted for a modular approach, separating concerns into individual components. This not only maintains clarity but also allows each component to evolve independently. For example, if we need to add new themes or adjust how the state is managed, it will be easier to scale.
- **State management**: We’re using React's `useState` hook within the `ThemeProvider` to manage the active theme, along with `useEffect` to apply the theme on mount. I considered using a more complex state management solution, but this keeps the feature lightweight and easy to integrate with the existing codebase.
- **Persistence**: The active theme is saved in local storage to persist across sessions. While it may be tempting to use a global state management library, this solution is sufficient for the needs of the theme system and avoids additional dependencies.

Open to feedback on the addation choice.

Feedback welcome!

### DIFF
--- a/src/features/ThemeProvider.js
+++ b/src/features/ThemeProvider.js
@@ -1,0 +1,18 @@
+// ThemeProvider component that provides theme configuration to the entire app.
+
+import React, { createContext, useContext, useState } from 'react';
+import { getCurrentTheme } from './themeConfig';
+
+const ThemeContext = createContext();
+
+export const ThemeProvider = ({ children }) => {
+  const [theme, setTheme] = useState(getCurrentTheme());
+
+  return (
+    <ThemeContext.Provider value={{ theme, setTheme }}>
+      {children}
+    </ThemeContext.Provider>
+  );
+};
+
+export const useTheme = () => useContext(ThemeContext);
--- a/src/features/ThemeSwitcher.js
+++ b/src/features/ThemeSwitcher.js
@@ -1,0 +1,22 @@
+// ThemeSwitcher component to allow users to switch between themes.
+
+import React from 'react';
+import { useTheme } from './ThemeProvider';
+import { setCustomTheme } from './themeConfig';
+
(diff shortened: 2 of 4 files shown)


robertobandini:

PR #1699, opened 2026-03-20
Allow skipping slow checks

marcovr:
Add cluster settings to allow skipping namespace authorization & metrics endpoint checks to speed up connecting to a cluster.

When connecting to a large-ish cluster, the namespace authorization check alone can take several seconds because it runs once per namespace. See the sample [logs.txt](https://github.com/user-attachments/files/26146578/logs.txt) or the following cropped image:

image

Similarly, the prometheus check also takes some time and keeps retrying despite no endpoint existing.
Because I often switch between multiple clusters, I would like to have the option to skip this check.


dex4er:
Generally I'm against any "hidden" settings. Either the setting is a part of the settings for cluster or for the application or we change the logic.

Perhaps we should just disable these checks at all or make them "lazy": so it should not start the startup and eventually be finished in the background.

Currently I don't use more complex RBAC so I'm not able to test significant changes if the checks are disabled for anyone. My guess is that the checks hide namespaces without access but anyway: we have a list of accessible namespaces in the settings already so extra checks seems to me not necessary.

marcovr:
> Generally I'm against any "hidden" settings. Either the setting is a part of the settings for cluster or for the application or we change the logic.

Fully Agree. I only did it this way because that's the easiest way to skip those steps without having to dig too deep into the codebase :-)
But I'll try to integrate it into the cluster settings

robertobandini:
Hi and thanks about your contribute @marcovr!

marcovr:
Did you get a chance to have a look at the PR?


This is how I integrated it into the cluster settings:


image


image

dex4er:
@claude [model:opus-4.8] Review PR and check if disabling this check for namespaces can break something (ie. list of namespaced will be missing, etc.) or we can get another regression.

dex4er:
@marcovr ah Claude is right. Now I remember with your PR I had all metrics just disabled.

marcovr:
> Now I remember with your PR I had all metrics just disabled.

Indeed, in a cluster which does provide metrics, this probably completely disables them. My focus was only on clusters which don't provide any metrics anyway. From that point of view, it really does just disable the endpoint detection.
But I guess my description is not very clear, as the option essentially just disables the whole metrics feature.

I'm also ok with removing that option and only keeping the checkbox for the namespace permissions. That would already significantly speed up the initial connection.

robertobandini:
Hi @marcovr! Sorry for the long silence on this one 🙏

Yes, let's go with your proposal: keep only the **namespace authorization checkbox** and drop the metrics one. The "disable metrics fetching" use case is tracked in #1703, better to handle it there with the right semantics.

I tested the namespace checkbox on a kind cluster with a restricted ServiceAccount (RBAC only in one namespace). With the checkbox on, the namespace list is unchanged, the sidebar shows all resource types and the forbidden ones just show an empty list with the usual warning icon, no errors or crashes. The setting is persisted correctly. So from our side it works as described!

FYI @dex4er opened #2250 which proposes an automatic short-circuit for the same per-namespace check. Your opt-in checkbox is a good complement in the meantime, especially with RBAC authorizers where the reviews are complete but slow on big clusters.

Could you please rebase on main (we moved to Vitest in the meantime) and remove the metrics option? Then we can merge. Thanks a lot for your patience!

marcovr:
@robertobandini no worries!
I rebased the PR and removed the metrics option. Also, thanks for linking the related issues

robertobandini:
Thanks @marcovr for the quick turnaround! The PR is now minimal and CI is green. We already did a manual check on a cluster with restricted RBAC (see the comment above), so I'm removing the "needs manual checks" label.

@dex4er @omarluq @MatiasRoje could you give this a final review when you have a moment? It's a small opt-in change (a single checkbox in the cluster settings). If there are no objections or reviews by Wednesday (September 2), we'll go ahead and merge it. Thanks a lot 🙏

robertobandini:
Hi @marcovr! The review window closed with no objections from the team, so this is going in. Thank you for the feature and for promptly reworking it after the feedback! 🙏

PR #1700, opened 2026-03-21
Add option to bypass Freelens internal KubeApi proxy

### LINKED ISSUE #766: Make k8s-lens-proxy optional (atleast  for port-forward and exec)

I would like to like to have the ability to disable the k8s-lens-proxy.

A side from "i don't know, why the proxy exists" in general, it causes trouble in context of pod shell and terminal.

Background: I have to reach a private Kubernetes Cluster which is accessible behind a SSH Bastion Jump host.

Instead working with local port forward, I'm using the ssh `-D` to setup an SOCKS5 proxy. The produces is described here: https://kubernetes.io/docs/tasks/extend-kubernetes/socks5-proxy-access-api/

But this is not compatible with `kubectl proxy` or with the lens proxy, because I got an error in Lens:

`error: unable to upgrade connection: error dialing backend: dial tcp: lookup ..eks.amazonaws.com: no such host`

Thats true, because the hostname is only resolvable from the bastion host, but not from the local host.

But invoking `kubectl port-forward` and `kubectl exec` outside of Lens works. I also guess there might be an bug in Kubernetes proxy function. 

But I see no reasons for using the proxy construct in general.

Ref: https://github.com/kubernetes/kubernetes/blob/122219790f82f5843d99aacbcb3d96162161e027/staging/src/k8s.io/client-go/tools/clientcmd/api/v1/types.go#L87

### Why is this needed?

_No response_

diodonfrost:
It fixes #766 

## Description

This PR adds a "Bypass Freelens Internal KubeApi Proxy" toggle in **Preferences > Proxy** so that users behind corporate proxies, bastion hosts (e.g. Teleport), or SOCKS5 tunnels that perform TLS certificate re-writing can use Freelens without broken connections.

When the toggle is enabled, kubectl and helm commands use the original kubeconfig directly instead of routing through the Freelens internal proxy. This was a supported feature in Lens that users migrating to Freelens have been requesting.

## What changed

* Added a new boolean user preference `bypassKubeApiProxy` (default: `false`).
* Added the preference toggle UI in Preferences > Proxy, following the same pattern as the existing "Allow untrusted certificates" toggle.
* Added persistence for the new preference (fromStore/toJSON in storage).
* Modified `KubeconfigManager.ensurePath()` to return the original kubeconfig path when bypass is enabled, skipping proxy kubeconfig generation.
* Updated snapshot tests to include the new preference block.

## Notes

* The toggle defaults to off, so existing behavior is unchanged for all users.
* When bypass is enabled, only kubectl/helm commands are affected. Cluster communications within Freelens still go through the internal proxy.
* The implementation follows the exact same DI and preference patterns as `allowUntrustedCAs` for consistency.

## Screenshot
screenshot-kubeapi-proxy-en


dex4er:
Settings for proxy in global preferences are for non-cluster traffic.

Settings for proxy for cluster are for cluster traffic.

Then: this PR has this setting in the wrong place.

Anyway, then it must be either mutual exclusive with "HTTP Proxy URL" setting, as this checkbox make the proxy in the app unused, or, better way: it should honor this setting using environment variables (HTTP_PROXY, HTTPS_PROXY).

diodonfrost:
Thanks for the review @dex4er , and for taking the time to look at this.

A bit of context on the design choices:

**On the placement**: I looked at the existing split in the codebase and the httpsProxy setting already exists at both levels — Preferences > Proxy for non-cluster traffic, and Cluster Settings > Proxy with the hint "Used for communicating with Kubernetes API." That matches exactly the rule you described, so I'll move the bypass toggle next to the per-cluster HTTP Proxy field in Cluster Settings > Proxy. That also makes the mutual-exclusivity UX easier since both controls end up in the same panel.

**One ergonomic concern I'd flag though**: bastion users (typically Teleport) route all their clusters through the same bastion, so a strictly per-cluster toggle means repeating the same setting for every imported cluster. Would you be open to an optional global default in Preferences > Kubernetes that new clusters inherit at import time (with the per-cluster setting remaining authoritative)? If you'd rather keep it strictly per-cluster, that's fine too — just wanted to float the hybrid option.

**On the HTTP_PROXY / HTTPS_PROXY suggestion**: Let me clarify the use case — I don't think it quite fits the forward-proxy model. The scenario isn't a forward HTTP proxy sitting between the user and the internet — it's a Kubernetes bastion like Teleport, which opens a local https://127.0.0.1:port endpoint per cluster (via tsh proxy kube) and writes these local endpoints directly into the kubeconfig, each protected by Teleport's own TLS certificate. In that setup:

- There is no HTTP forward proxy to delegate to, so HTTPS_PROXY wouldn't apply — the kubeconfig already points at the bastion's local endpoint.
- The symptom I'm seeing is: plain REST calls work fine through Freelens's internal proxy, but streaming endpoints (kubectl exec, port-forward, logs -f) fail with a TLS error. That's consistent with the internal proxy's handling of the SPDY/WebSocket upgrade when the upstream presents the bastion's own cert chain, though I haven't pinpointed the exact line — what I can confirm is that bypassing the internal proxy makes the symptom disappear. That's the same issue reported in #766 .

Bypassing the internal proxy entirely is the only reliable way I've found to make these streaming operations work through a bastion that rewrites certificates — which is why Lens shipped it as an explicit toggle rather than relying on env vars.

**On mutual exclusivity with the HTTP Proxy URL field**: good point, I agree. With the toggle moving to Cluster Settings > Proxy, both controls will live in the same panel — I can disable/grey out the HTTP Proxy URL input when the bypass is enabled, or surface a warning. Let me know which UX you prefer.

Happy to iterate on the placement and UX details — just want to make sure we don't drop the core fix, since env vars won't cover the bastion case.

dex4er:
Ok, so I would agree this option might be just in the global settings, to avoid setting it separately for each cluster in the same network.

However, the implementation of this feature just skips creating local $KUBECONFIG (ie. `/var/folders/pf/4zp8d_xs76s8r0hl16qzpshm0000gp/T/kubeconfig-c120532290e2862cca7b34792e3b4d8b` in my system) that contains something like:

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: >-
      LS0AAAAAAAAAA==
    insecure-skip-tls-verify: false
    server: https://127.0.0.1:59166/c120532290e2862cca7b34792e3b4d8b
  name: kind-kind
contexts:
- context:
    cluster: kind-kind
    user: proxy
  name: kind-kind
current-context: kind-kind
kind: Config
preferences: {}
users:
- name: proxy
  user:
    password: fake
    username: lens
```

Instead, the original ~/.kube/config file is used.

it means that for each cluster I connect I've got the same context. Ie. if I have kind1 and kind2 cluster then in each terminal for separate cluster I have the same "kind1" or "kind2" context (as it is the default context) then the terminal runs kubectl and helm on INCORRECT cluster! This is fatal regression and it renders this feature dangerous to use.

I think the temporary $KUBECONFIG file at least should copy the original $KUBECONFIG file and switch the context to the same as for the connected cluster.

diodonfrost:
Thanks for catching this @dex4er, you're right, returning the original kubeconfig as-is is a fatal regression. Just pushed a fix.

dex4er:
@claude [model:fable] Check if PR brings regression, ie. the environment in terminal is incorrect and bypassing freelens-k8s-proxy would mean that original ~/.kube/config is used then the context is shared betwen cluster views.

dex4er:
@claude [model:fable] Actually, why I would need to bypass freelens-k8s-proxy as, ie. I currently can use HTTP proxy or even Proxyman with TLS cert (as TLS man-in-the-middle proxy) and then eventually I need to set `allowUntrustedCAs` to accept non-original certificate?

What exactly use case this PR tries to fix and if there is a better way to handle it?

What is the final recommendation about this PR?

jkroepke:
Additionally, I using Lens with AWS auth against EKS auth. On each AWS session renew, with AWS SSO Login, Lens somehow stucks and doesn't use the new token from exec credential...

diodonfrost:
@dex4er I pushed `ca11492f` addressing the review feedback. The core change: `createBypassKubeconfig` no longer goes through the client-node parser + `dumpConfigYaml` (which is what dropped fields and froze tokens). It now round-trips the **raw** kubeconfig YAML, so every field is preserved verbatim.

Mapping to the points raised:

1. **Lossy copy** — fixed. `proxy-url`, `tls-server-name`, impersonation (`as`) and `extensions` are now preserved. This was the critical one for the SOCKS5 / #766 case, where the copied kubeconfig was losing the very `proxy-url` that makes it work. New test: *"preserves fields that a lossy copy would drop (proxy-url, tls-server-name, impersonation, exec)"*.
2. **Relative paths** — fixed. Relative `certificate-authority`, `client-certificate`, `client-key` and `token-file` paths are made absolute against the original kubeconfig's directory before the temp copy is written, so they no longer resolve against the temp dir. New test covers this.
3. **`token-file` inlined/frozen** — fixed. It stays a live reference now, so short-lived rotating tokens keep working instead of being snapshotted at load time. Verified by test.
4. **Per-cluster `defaultNamespace` ignored** — fixed. It's applied to the pinned context.
5. **Staleness** — the temp copy is created on (re)connect; I reworded the hint to say *"Takes effect on reconnect."* It isn't regenerated mid-session — let me know if you'd rather it be rebuilt on every `ensurePath()`.
6. **Scope wider than the hint claimed** — fixed. The hint now states the bypass also affects in-app Helm releases and the node shell, and that the cluster resource views are unaffected.

`current-context` is still pinned per-cluster and each terminal exports its own `kubeconfig-<clusterId>`, so the earlier fatal shared-context regression stays fixed (still test-covered).

Two open points for you:

- **`exec.command`**: I deliberately left it verbatim rather than absolutizing it, since it's normally a binary resolved via `PATH` (`tsh`, `aws`, …) and making it absolute could break it. Happy to change if you'd prefer otherwise.
- **Mutual-exclusivity with the HTTP Proxy URL field**: since we settled on keeping this as a single global setting (not per-cluster), it no longer sits next to a per-cluster proxy field. Let me know if you still want any interlock/greying-out.

Re-requesting your review when you have a moment.

diodonfrost:
@jkroepke on the AWS SSO / EKS exec-credential refresh: with this bypass enabled, kubectl/helm run against your original kubeconfig and re-invoke the `exec` credential plugin on each call, so a renewed AWS SSO session should be picked up without reconnecting — the bypass copy keeps `exec` as a live reference rather than inlining a token. The "stuck on session renewal" behaviour you're seeing sounds like it lives in the internal-proxy path, which is a separate issue from this toggle and probably worth tracking on its own.

jkroepke:
> plugin on each call

This would differ to manual kubectl calls. [exec credentials](https://kubernetes.io/docs/reference/config-api/client-authentication.v1beta1/#client-authentication-k8s-io-v1beta1-ExecCredentialStatus) return a `expirationTimestamp` and the binary is not invoked until the timestamp hits.

diodonfrost:
You're right, and I was imprecise — thanks for the correction.

Within a single long-running client-go process, the exec credential is cached and the plugin isn't re-invoked until its `expirationTimestamp` is reached (or a 401 forces a refresh). So "on each call" is wrong for that case.

The distinction that matters here is the execution context, because the two don't share a cache:

- **Separate `kubectl`/`helm` CLI invocations** (what this toggle affects): each command is a fresh process with a cold in-memory cache, so it re-runs the exec plugin per command and uses your current AWS SSO session. `aws eks get-token` just re-signs locally from `~/.aws/sso/cache`. This is the same behavior as the manual `kubectl` you already rely on.
- **A long-running client-go process** (Freelens's in-app views / internal proxy): one authenticator, credential cached against `expirationTimestamp` exactly as you describe.

With the bypass, the temp kubeconfig keeps `exec` as a live reference (no inlined/frozen token), so terminal `kubectl`/`helm` pick up a renewed SSO session on the next command, without reconnecting Freelens.

The "Lens stucks and doesn't use the new token on SSO renewal" symptom you hit is the second context — the long-running internal path, not this toggle. That's a separate issue worth tracking on its own; this PR doesn't touch that path.

jkroepke:
Is this ChatGPT now?

PR #1750, opened 2026-03-30
Feat: pin/save searches

### LINKED ISSUE #939: Pin/Save Searches

I would love the ability to pin or save my searches.  This way, if desired, I can look at different resource types without having to re-enter my search every time.

This feature is implemented in the "commercial" version of Lens and is very useful.

### Why is this needed?

When I'm working in a namespace(or even multiple namespaces at once) that have a lot resources in it.  I often do a search to narrow down the list to a certain set.  For example, I want to see all the PODs for "app1".  However, when I switch to look at the Secrets for that same app, I have to re-enter that "app1" search term.  Additionally, if I want to switch back to PODs, I have to re-search again.

Pinning a search term to use across multiple resource listings would be extremely useful.  At the very least, keeping that search term "saved" would be useful too.  That way, maybe I could search for one term for PODs, re-enter that term for Secrets, but when I go back to PODs, it is still showing me the search results from before.

maifeeulasad:
closes #939 

pin and save searches feature for freelens implemented by Maifee Ul Asad


dex4er:
@maifeeulasad Generally it looks interesting but in the PR files changed I see a lot of `searchhistory="[object Object]"` and it doesn't look correct for me.

maifeeulasad:
@dex4er thanks for noticing! 

here is the thing, 
 - i made changes to `packages/core/src/renderer/components/input/input.tsx`. where i added this `searchHistory` props, which was being passed down to input. 
 - and to make the ci happy, i updated the snaps. and that resulted in that `searchhistory="[object Object]"` in all over places
  
  
  
 but now it is resolved, as i have excluded this as well from passing down to native elements.

here is a sneak peek of the solution:
```
render() {
    ...
    const {
      ...
      id,
      dirty: _dirty, // excluded from passing to input-element
      trim,
      blurOnEnter,
      searchHistory: _searchHistory, // excluded from passing to input-element
      ...inputProps
    } = this.props;
```

care to take a look into it again please?

robertobandini:
Hi @maifeeulasad, thanks for fixing the snapshot issue from @dex4er's review, and sorry for the late reply.

This won't go into v1. The v2.0.0 milestone will regenerate all snapshots and change part of the tooling, so we'll park the PR as it is and rebase it only after v2 lands on main. Please don't rebase before that, it would be wasted effort. I'm updating the labels accordingly.

In the meantime, the open design question is how pinned/saved searches should interact with the persistent and linked search modes (v1.6.2, v1.10.0). Please post your proposal on #939, so the discussion and the community feedback stay on the issue.

As with every feature, the direction is decided by community demand. Building consensus around #939 is the best way to give this PR a clear path after v2.

Thanks for your patience.

PR #1773, opened 2026-04-07
feat: change accent color to Lens blue (#3d90ce) and add macOS build …

lumi-aamir:
…workflow

Replace default teal (#00a7a0) with Lens blue (#3d90ce) across themes, charts, splash screens, and components. Add manual GitHub Actions workflow to build macOS arm64 DMG as downloadable artifact.


lumi-aamir:
ignore. created pr by mistake to wrong base

PR #1786, opened 2026-04-20
Use GET instead of POST for Prometheus query_range requests

### LINKED ISSUE #1785: Use GET instead of POST for Prometheus query_range to support read-only RBAC roles

Change the HTTP method from POST to GET for Prometheus `query_range` and `query` requests in `packages/core/src/main/get-metrics.injectable.ts`.

Currently the metrics fetcher sends a POST request through the Kubernetes service proxy:
```
POST /api/v1/namespaces/.../services/prometheus-operated:9090/proxy/api/v1/query_range
```

This should be changed to:
```
GET /api/v1/namespaces/.../services/prometheus-operated:9090/proxy/api/v1/query_range?query=...&start=...&end=...&step=...
```

Prometheus supports both GET and POST for these endpoints and returns identical results.

### Why is this needed?

On Azure AKS with "Azure RBAC for Kubernetes Authorization", HTTP POST to a service proxy is mapped to a **write** action (`services/proxy create`), while GET is mapped to a **read** action. This means users with only the built-in "Azure Kubernetes Service RBAC Reader" role get HTTP 403 on the POST request:

```
services "prometheus-operated:9090" is forbidden: User cannot create resource "services/proxy"
```

The metrics show as "Metrics not available" even though Prometheus is correctly discovered via the Operator provider and GET requests to the same proxy path succeed.

Azure does not offer a granular DataAction to allow proxy-POST without granting full service write access — the only fix is to use GET instead of POST.

This likely also affects GKE and EKS setups where read-only RBAC roles are used, since standard Kubernetes RBAC also distinguishes the `proxy` sub-resource verb (`get` vs `create`) based on HTTP method.

**Tested:** GET and POST to `/api/v1/query_range` return byte-identical responses from Prometheus.

joggeli34:
Azure RBAC maps HTTP POST through the Kubernetes service proxy as a create/write action on services/proxy. This causes 403 errors for users with read-only roles (e.g. Azure Kubernetes Service RBAC Reader) when Freelens tries to fetch metrics via the Kubernetes API proxy.

The Prometheus query_range API supports both GET and POST with identical results. Switching to GET ensures metrics work for users with only read permissions on the cluster.

Fixes #1785


dex4er:
I think we can replace POST with GET, however there is a risk to overflow limit for the query string size but I think we still have a safe limit.

robertobandini:
Thanks about your contribute @joggeli34!

PR #1809, opened 2026-04-23
Smoother log search pattern updates

ErionBajrami:
The search input previously triggered a full regex scan over every log line on each keystroke, causing noticeable lag on large log buffers. Debounce the scan by 250ms and keep the input controlled locally so typing stays smooth. Enter and Ctrl/Cmd+F selection flush immediately.


ErionBajrami:
Closes #1810

ErionBajrami:
Forgot to run tests. Rookie* mistake!

Made some changes to fix the arrow buttons on next overlay and handled the grayed out button when no results are there. Will do some ad-hoc testing on monday.

ErionBajrami:
Tested ad-hoc style :)

dex4er:
@ErionBajrami LGTM. Thanks for patch!

PR #1825, opened 2026-04-29
Custom URL for kubectl download mirror 

### LINKED ISSUE #1823: Add support for custom kubectl binary download mirror URL

Add an optional preference setting to specify a custom base URL for downloading kubectl binaries (in addition to the existing Default (Google) and China (Azure) mirror options).

### Why is this needed?

Many organizations run in environments where direct Internet access is restricted by firewall policy. This will allow Freelens to download the correct kubectl binaries in restricted environments.

ErionBajrami:
Adds a "Custom" option to the kubectl download mirror dropdown in Preferences → Kubernetes → Kubectl. When selected, a "Custom mirror URL" input appears where the user enters a base URL. That URL replaces the built-in mirror in the kubectl download path.

Can
- Use any HTTPS (or HTTP) base URL serving the standard kubectl release layout
- Works on macOS, Windows, and Linux (unlike the China mirror which is win32/linux only)
- Trailing slashes in the input are stripped automatically
- Falls back to the default mirror if the custom field is left empty
- URL is validated as a URL on input

Cannot
 - Mirror must follow the upstream dl.k8s.io/release directory layout — Freelens appends /v{version}/bin/{platform}/{arch}/kubectl[.exe]. Mirrors with a different path scheme won't work.
- No per-platform or per-architecture override (one URL for all)
- No authentication support (no headers, tokens, basic auth, or client certs)
- No checksum/signature verification beyond what already exists for the default flow
- Doesn't proxy or cache — every download hits the URL you provide
- Only affects kubectl. Other downloads (Helm, etc.) are unchanged.

This should do the trick.

+ Also did very small refactor for China, download mirror

Closes #1823

image

image


ErionBajrami:
For some reason i feel like it's missing something, im just no quite sure what it is.

Would organizations even follow the directory layout for their kubectl binary downloads?

robertobandini:
Thanks @ErionBajrami!
What do you think @inquinity ?

inquinity:
Thanks @ErionBajrami, @robertobandini 
I am going to take a look. I actually just finished building the feature myself and was testing it for usability; I'm going to switch to this branch.
It looks like authentication might also be an issue. What are your thoughts on handling authentication values? I don't want to store anything, but I'm not sure if using exported values is a good-enough solution.

inquinity:
My version of the feature implementation has in-place editing. Which UI do you prefer? I can add it as a PR, but I don't want to add conflicting PRs unless you want to see it.
image

robertobandini:
@inquinity Thank you about working also to this yourself!

Maybe personally at first sight I prefer the UI version made by @ErionBajrami but it's just only my personal opinion.
Generally I prefer that the community choose every UI decision in a democratic way.
In case we can make a poll about this on the discussions section of the repo.

About the related auth topic, could be that currently you are the only one that did this request so we can start from what is more pragmatic about your case.

I also tag the release engineering team @dex4er @MatiasRoje @omarluq so we can listen also to their opinion about this.

Thanks guys about your contribute!

inquinity:
@robertobandini I just wanted to hear your opinion on authentication. I don't think I want to encourage this or have Freelens be responsible for (safely) storing the secret or risking someone leaking it in logs.

I say we leave it out of the current feature.

inquinity:
@robertobandini I would like to add some things to this branch/PR, specifically unit tests and rebase to Freelens 1.9.0. What is the workflow? Do we need to merge this to main or can we merge it to a new branch and update it there?

inquinity:
@robertobandini @ErionBajrami Can we update this PR? I have a new branch that adds unit tests and imports the Freelens 1.9 release changes.

dex4er:
Let's see if it passes the tests with the current main branch then we can merge it.

ErionBajrami:
@robertobandini @inquinity Sorry everyone, apologies for the lack of inactivity. I've been tied up this past month with something personal.

robertobandini:
@ErionBajrami Hi! No problems at all, on the contrary: thank you all about your contribute!

PR #1838, opened 2026-05-04
Add correct formatting for IPv6:Port tuple

YannikSc:
This implements proper IPv6:Port tuple formatting for the Endpoints listing, making it much easier to read them in the listing. IPv4 endpoints are still displayed as usual. 

I could not find any other places which show IP:Port tuples but if anyone knows of places I should check feel free to mention it.


dex4er:
Thanks!

PR #1847, opened 2026-05-06
Implement dynamic custom json themes extension loader

serfersac:
Addresses #550. Introduces a modular custom themes loader engine that safely scans the application userData directory for user-contributed theme JSON specifications. Validates the structural schema attributes on startup and injects valid parameters directly into the core theme registry Map, allowing custom user colors to natively propagate through the preference selection options without introducing UI bloat or form debt.

### DIFF
--- a/packages/core/src/renderer/themes/custom-theme-manager.injectable.ts
+++ b/packages/core/src/renderer/themes/custom-theme-manager.injectable.ts
@@ -1,0 +1,59 @@
+
+
+/**
+ * Copyright (c) Freelens Authors. All rights reserved.
+ * Licensed under MIT License. See LICENSE in root directory for more information.
+ */
+
+import { getInjectable } from "@ogre-tools/injectable";
+import { app } from "electron";
+import fs from "fs";
+import path from "path";
+import type { LensTheme } from "./lens-theme";
+
+const customThemeManagerInjectable = getInjectable({
+  id: "custom-theme-manager",
+  instantiate: () => {
+    const saveCustomTheme = (theme: LensTheme): boolean => {
+      const themesDir = path.join(app.getPath("userData"), "themes");
+
+      try {
+        if (!fs.existsSync(themesDir)) {
+          fs.mkdirSync(themesDir, { recursive: true });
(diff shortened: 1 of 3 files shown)


dex4er:
@serfersac I think it might be the best implementation of the custom themes I ever seen here. I think we still miss some presentation in GUI for it? But in my opinion it is the right direction we can go.

dex4er:
@serfersac I think we could have ie. the dialog for YAML / JSON editor or just file picker. Or even both.

dex4er:
@claude Review the PR. Check for regressions. See if we can easily add preferences for adding/removing/editing custom themes later.

robertobandini:
Hi @serfersac and thank you again for this work. As @dex4er wrote, this looks
like the right direction for the full custom themes system.

As I wrote on #1280, we plan to finalize the accent color feature (#1431)
right after the Freelens v2 release. So I want to share how we see the next
steps for this PR too, and ask the opinion of the other maintainers and the
community.

To be clear, this PR is not mergeable in its current state: the review above
found a critical issue (the new injectables are not registered, so the app
would crash at startup) and the GUI part is still missing. But even once
fixed, we would not merge it before v2 anyway. Freelens v2 changes the
foundations (ESM and more), and a custom theme file format is a compatibility
promise: once users start creating theme files, we cannot easily change the
schema anymore. So we want to design it once, and well, on top of v2.

The plan we propose:

1. First, right after the v2 release, we conclude #1431 (accent color, the
   #1280 bounty), which is a small and reversible preference.
2. Then we take this PR as the base and reference for the full custom themes
   system, fixing the points from the review (DI registration, schema
   validation, etc.) and adding the GUI part.
3. One more idea: with the extensions marketplace coming on v2, themes could
   become installable artifacts, distributed and curated like extensions,
   instead of plain files in the userData folder.

@dex4er, you proposed this direction first: do you agree with this plan?
@serfersac, would you like to stay involved and continue on this when we
start? Feedback from everyone is welcome.

FYI @mariomamo @leo-capvano @MatiasRoje

PR #1873, opened 2026-05-15
feat: add custom accent color preference

helh1723-lang:
## Summary
- Add a user preference for a custom accent color in Application > Theme.
- Persist and validate the custom accent color in the user preferences store.
- Re-apply the active theme whenever the accent color changes and override core accent CSS variables such as primary, textColorAccent, buttonPrimaryBackground, colorInfo, helmStableRepo, and menuActiveBackground.
- Reset the custom accent color when the theme is reset and cover it in the user-store test.

Closes #1280

## Validation
- git diff --check

## Notes
- I could not run the full pnpm test/build suite in this local environment because pnpm is not installed and corepack is blocked from writing its package-manager cache here.


robertobandini:

PR #1876, opened 2026-05-15
Avoid stack overflow when spreading large API resource arrays

agarkoff:
## Problem

On clusters with many installed components (Istio, Kyverno, Flux, KEDA, Velero, Cilium, cert-manager, etc.), `refreshAccessibility` crashes with:

```
RangeError: Maximum call stack size exceeded
at Object.requestApiResources (main.js:21189:27)
```

## Root cause

`requestApiResources` uses the spread operator to accumulate results:

```typescript
groupLists.push(...result.response);
// ...
resources.push(...result.response);
```

When using `push(...array)`, JavaScript passes every element as a separate function argument on the call stack. V8 has a hard limit (~125k arguments). On clusters with 100+ API resources across many groups, this limit is exceeded and the engine throws `RangeError: Maximum call stack size exceeded`, making the cluster permanently inaccessible in Freelens.

## Fix

Replace both spread-into-push calls with explicit `for-of` loops that push items one at a time, bypassing the argument stack limit entirely.

## Testing

Reproduced and verified on a cluster with 184 API resources across 66 API groups. The cluster loads successfully after the fix.


dex4er:
@claude What is performance degradation if we use `for (const item of result.response) { groupLists.push(item); }` instead of pushing elements in one call?

robertobandini:
Hi @agarkoff and thank you very much about your interest in Freelens and about your contribute!

PR #1913, opened 2026-05-29
Stop using Pod Ready=False as a proxy for pod issues

robertdyr:
## Summary

This removes `Ready=False` as a direct proxy for `Pod.hasIssues()`.
The Kubernetes `Ready` condition describes whether a pod is ready to serve traffic. That is not always the same thing as whether the pod has an issue a user should be warned about. `Ready=False` can mean several different things, including unready containers, readiness gates, missing statuses, terminal pod state, or containers that completed successfully.
This distinction matters for pipeline-style pods and Jobs. Pipeline step containers can terminate successfully with `exitCode: 0`, and succeeded Job pods are also reported as `Ready=False` after completion. In both cases, users do not expect Freelens to keep showing a warning for work that finished successfully.

## What Changed

`Pod.hasIssues()` now checks the concrete signals behind readiness instead of treating `Ready=False` itself as an issue.
The checks follow the same inputs kubelet uses to derive pod readiness as closely as possible ([kubelet Pod Ready condition logic](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/generate.go#L126)):

- regular container readiness;
- restartable init container readiness;
- readiness gates.

Freelens then applies issue-specific exceptions for cases that are known not to be user-facing problems:
- successfully terminated containers with `exitCode: 0` are not treated as issues;
- `Succeeded` pods are not treated as issues only because they are no longer `Running`.
-
`CrashLoopBackOff` is still treated as an issue, including for ephemeral containers. Pod phase remains a final fallback for other non-running/non-succeeded states.

### DIFF
--- a/packages/kube-object/src/specifics/pod.test.ts
+++ b/packages/kube-object/src/specifics/pod.test.ts
@@ -551,5 +551,22 @@ describe("Pods", () => {
 
       expect(pod.hasIssues()).toStrictEqual(false);
     });
+
+    it("should return false if a succeeded pod has a readiness gate that is not ready", () => {
+      const pod = getDummyPod({ running: 1 });
+
+      assert(pod.status);
+
+      pod.status.phase = "Succeeded";
+      pod.spec.readinessGates = [{ conditionType: "example.com/ready" }];
+      pod.status.conditions.push({
+        type: "example.com/ready",
+        status: "False",
+        lastProbeTime: 1,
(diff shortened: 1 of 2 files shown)


kavvadiasa (approved):

robertdyr:
@MatiasRoje think im done now.

dex4er:
Thanks!

PR #1914, opened 2026-05-31
Add custom theme color overrides

Gwani-28:
## Summary
- add persistent custom color overrides to user preferences
- merge valid color overrides into the active theme before CSS variables are applied
- expose a Custom colors panel in Application > Theme with color pickers and reset controls

Closes #1280

## Testing
- pnpm biome:check
- pnpm build:core
- pnpm build
- pnpm build:resources


robertobandini:

PR #1923, opened 2026-06-03
Prune unused native binaries to prevent warnings about Intel-based apps in MacOS

justincorrigible:
### Context
Users in the latest version of MacOS may be receiving this alert notification
image
and the likely cause (at least as of the current release, v1.9.0 from May 5, 2026) is that the macOS arm64 .pkg includes stray x86_64 Mach-O files inside production dependencies:
```
node-pty/prebuilds/darwin-x64/*
pnpm/dist/reflink.darwin-x64-*.node
```

The hook added in this PR prunes unused `node-pty` prebuilds and `pnpm reflink` native binaries for non-target architectures before signing/notarization.

### Verification:

- Confirmed the downloaded arm64 package checksum matches the release metadata.
- Ran the hook against the expanded arm64 package.
- Re-ran file across executable/native payloads; no x86_64 Mach-O files remained.
- Ran `pnpm biome check freelens/electron-builder.yml freelens/build/prune-unused-native-binaries.js`; Biome passed for the JS hook. It reported one checked file, so YAML is effectively not covered by that check.
- Did not run a full Electron package build locally.

Hope this helps!


dex4er:
@claude Please review

dex4er:
@claude fix issue 1 from review

dex4er:
@claude push changes (workflow unblocked them)

PR #1932, opened 2026-06-09
feat(pods): add Running Time column

ashishmistry36:
This PR adds a Running Time column to the Pods list table which calculates the real-time running duration of a Pod based on its oldest container's startedAt status.

### DIFF
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -196,7 +196,7 @@
     "webpack-node-externals": "^3.0.0"
   },
   "engines": {
-    "node": "^24.0.0"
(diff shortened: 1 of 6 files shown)


dex4er:
It is interesting information that we can present but currently it'll duplicate Age column in most cases. The pod view is already overloaded with informations then I think this new column should be hidden by default and optionally enabled.

image

ashishmistry36:
Yeah that makes sense,  I've updated this PR to hide the Running Time column by default.

robertobandini:
Hey @ashishmistry36 thank you about your contribute!

PR #1940, opened 2026-06-11
Add pods, instance type, node group, capacity columns and requests overlay to nodes list

bandoche:
## What

Extends the cluster Nodes list with information that is essential when operating managed-cloud clusters (EKS, GKE, AKS), where pod capacity per node is limited and nodes come from different node groups / capacity types.

### New columns (hidden by default)

All four new columns are **hidden by default** (the nodes table already has many columns), and can be enabled via the configurable-columns menu. They are also sortable and searchable. Nodes without the relevant labels show empty cells.

| Column | Source | Notes |
|---|---|---|
| **Pods** | non-terminated pods on the node vs `status.allocatable.pods` | e.g. `45 / 58`, with a usage bar — on EKS the allocatable pod count is limited by ENI capacity, so this shows scheduling headroom at a glance |
| **Instance Type** | `node.kubernetes.io/instance-type` (fallback: legacy `beta.kubernetes.io/instance-type`) | e.g. `c6a.8xlarge` |
| **Node Group** | `eks.amazonaws.com/nodegroup` → `karpenter.sh/nodepool` → `karpenter.k8s.aws/ec2nodeclass` → `cloud.google.com/gke-nodepool` → `kubernetes.azure.com/agentpool` → `agentpool` | EKS managed node groups, Karpenter, GKE, AKS |
| **Capacity** | Karpenter / EKS / GKE (`gke-provisioning`, `gke-spot`, `gke-preemptible`) / AKS (`scalesetpriority`) | normalized to `spot` / `on-demand` / `preemptible` across clouds |

### CPU / Memory usage vs requests

- The CPU and Memory cells now show `actual usage / requests` as text under the bar (usage from metrics-server, requests summed from the pod specs of non-terminated pods on the node).
- `LineProgress` gains an optional `secondaryValue`, rendered as a lighter (0.5 opacity) line behind the main one. The nodes list uses it to show requests behind usage, so a node whose requests already fill its allocatable capacity is visible at a glance even when actual usage is low.
- Bars fall back to metrics-server usage vs node allocatable when Prometheus metrics are unavailable, and the requests line is shown even when no usage source responds (requests come from pod specs and need no metrics).
- Node list bars are 5px tall for readability (other LineProgress consumers keep the 3px default).
- Fixes CPU usage text being formatted with byte units (`0.0Ki`) — cores are now formatted as numbers.

## Addressed review feedback

- **Hidden by default**: the four new columns no longer expand the default table view; they are opt-in via the column menu (same mechanism the Pods and Deployments lists already use).
- **GKE & AKS support**: Node Group and Capacity now read the GKE (`cloud.google.com/gke-*`) and AKS (`kubernetes.azure.com/*`, legacy `agentpool`) labels in addition to EKS/Karpenter, with values normalized across clouds.

## Verification

- `tsc --noEmit` on `packages/core` passes; biome clean; node-related and LineProgress-consumer unit tests pass
- Verified end-to-end on an EKS 1.35 cluster (managed node groups + Karpenter, spot and on-demand): all columns populate when enabled, the requests overlay renders behind usage, the no-metrics-server case still shows the requests bar with `N/A` usage text, and the new columns start hidden on a fresh profile


dex4er:
@bandoche
 It is great feature but I could approve it when:

1. the new colums are hidden by default as currently we have too many columns already and adding more will break the view:

image

2. it'll support also GKE and AKS, ie `cloud.google.com/gke-nodepool`, `cloud.google.com/gke-provisioning` etc.

dex4er:
@claude Write what annotations/labels can be used for GKE and AKS to use for this PR

bandoche:
@dex4er Thanks for the review! I've addressed both points and rebased onto the latest `main`:

1. **Hidden by default** — the four new columns (Pods, Instance Type, Node Group, Capacity) are now passed via `defaultHiddenTableColumns`, so they don't expand the default table view and are opt-in through the column menu (same approach the Pods and Deployments lists already use).

2. **GKE & AKS support** — using the labels from your list:
   - Node Group also reads `cloud.google.com/gke-nodepool` (GKE), `kubernetes.azure.com/agentpool` and legacy `agentpool` (AKS).
   - Capacity also reads `cloud.google.com/gke-provisioning`, `gke-spot`, `gke-preemptible` (GKE) and `kubernetes.azure.com/scalesetpriority` (AKS), normalized to `spot` / `on-demand` / `preemptible` across clouds.

Verified on an EKS 1.35 cluster (managed node groups + Karpenter, spot and on-demand) that the columns start hidden on a fresh profile and populate correctly once enabled.

bandoche:
Windows job failed at the Electron-distribution unpacking step (exit 127) before any of this PR's renderer code is involved; main passes on Windows, so this looks like a flaky runner failure. Could you re-run the Windows job?

dex4er:
@claude Rebase or merge main then fix the confict in a new PR

robertobandini:
Thank you very much about your contributes @bandoche!

Issue #551, opened 2025-04-07
Folder sync only triggers at app start


Cenness:
### Describe the bug

Kubeconfig folder isn't synced after app started

### Reproduction steps

1. Make a folder with kubeconfigs, like `%appdata%\Lens\kubeconfigs`
2. Start freelens with no folders to sync set up
3. Add folder to sync
4. Wait
5. No configs in catalog
6. Restart app
7. All configs from folder are present

### Expected behavior

Configs from sync folder added as soon as folder added to app config.  
New config added to catalog as soon as new file added to previously added sync folder.

### Freelens version

1.2.0

### Operating system type and version

Windows 10 [Version 10.0.19045.5608]

### Installation method

MSI

ngowuys:
I am facing with same problem

robertobandini:
Thanks @Cenness and @ngowuys!
If anyone could help with this, just let us know!

lmhinnel:
I dont know how to fix ít, but you can shut down the app and start it again :v

FredericLeroy:
Even this doesn't work for me, the only way to have it working is to use "Add from kubeconfig" and copy paste kubeconfig content

joshsmithers:
> Even this doesn't work for me, the only way to have it working is to use "Add from kubeconfig" and copy paste kubeconfig content

If anyone is having this issue on MacOS, check the activity monitor and make sure that FreeLens is completely stopped. Then open it again and your new folder(s) should be synced.

robertobandini:
Are you still having this problem?
@dex4er Are you experiencing this problem?

PR #1941, opened 2026-06-11
Propagate changes for kubeconfig entries

### LINKED ISSUE #1108: Kubeconfig Syncs is not working as expected.
### Describe the bug

Both Sync File(s) and Sync Folder(s) options under Kubeconfig Syncs settings are not working as expected. I added both files and folder where my Kubernetes configs files exists, but freelens does not show clusters from those files.

### Reproduction steps

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Freelens version

1.5.3

Windows 11 Pro

### Installation method

MSI

_No response_

### Logs

_No response_

### Kubeconfig

_No response_

### Additional context

_No response_

yehuda-kahan:
## Summary

Fixes [#1108](https://github.com/freelensapp/freelens/issues/1108) and [#551](https://github.com/freelensapp/freelens/issues/551).

Adding a kubeconfig file or folder via **Preferences → Kubernetes → Kubeconfig Syncs** does not start a watcher; the clusters never appear in the catalog. Removing an entry has the same problem — the watcher keeps running. Restarting the app (force-quit on macOS, kill via Task Manager on Windows) makes the same entries work because preferences are reloaded fresh on startup.

Reported on Windows (#1108, #551) and macOS (#551 [comment by @joshsmithers](https://github.com/freelensapp/freelens/issues/551#issuecomment-3165466863)). The bug is in shared, cross-platform code.

## Root cause

Two-layered problem in the user-preferences IPC pipeline interacting with `KubeconfigSyncManager`:

**Layer 1 — descriptor identity is not preserved across IPC reloads.** The `syncKubeconfigEntries` descriptor's `fromStore` was called on every IPC sync (not just at startup) and created a fresh `observable.map` each time. Anything that captured a reference to the original map — including the `kubeconfigSyncs` injectable, which is evaluated once and feeds `KubeconfigSyncManager` — was now pointing at a stale map. The renderer's edits landed on a new map instance the main process was no longer observing.

**Layer 2 — the manager's runtime hookup couldn't tolerate replacement.** `KubeconfigSyncManager.startSync()` used `observe(map, change)` to react to add/delete deltas. `observe()` fires for in-place mutations on the observed map but cannot fire when the entire map is swapped out from underneath it.

Together, this is why a restart "works": on next boot, preferences are read fresh into a brand-new map, and `startSync()` synchronously walks `keys()` once on startup — so the entries you added in the previous session appear correctly.

## Fix

**`preference-descriptors.injectable.ts`** — `syncKubeconfigEntries` now owns a single `ObservableMap` in a closure. `fromStore` mutates it in place (delete missing keys, set new/changed keys, return the same instance). Identity is preserved across IPC reloads, so all consumers stay in sync.

**`manager.ts`** — `observe(map, ...)` is replaced with a `reaction(() => Array.from(map.keys()), reconcile)`. The reaction recomputes the desired-paths set and diffs it against currently-watched paths, starting/stopping watchers accordingly. The `directoryForKubeConfigs` is kept in the desired set unconditionally so it is never accidentally stopped. Robust whether the underlying map is mutated in place or replaced wholesale.

Either layer alone is sufficient to fix the bug; both together is defence-in-depth — the descriptor change makes the contract right, and the manager change makes the consumer resilient.

## Tests

Added 9 unit tests in `kubeconfig-sync.test.ts`:

- **runtime add via manager** — `localKubeconfigSyncs.set(path, ...)` after `startSync()` results in `watchKubeconfigFileChanges(path)` being called.
- **runtime delete via manager** — `localKubeconfigSyncs.delete(path)` after `startSync()` calls the watcher's disposer.
- **`directoryForKubeConfigs` is always watched** — even when absent from the map, even when explicitly removed from it.
- **descriptor IPC-replay path** — two consecutive `fromStore()` calls return the same `ObservableMap` instance (`Object.is`), and entries added/removed via `fromStore()` propagate to the manager's watchers.

All 22 tests in the file pass (13 pre-existing + 9 new).

## Validation

Reproduced the bug and verified the fix on Linux/WSL using a locally-built AppImage that mirrors the upstream electron-builder pipeline (same Linux build steps as the published macOS/Windows artifacts). The fix lives in cross-platform code (mobx + the shared descriptor pattern) with no `process.platform` branching, so the same code path runs on every supported OS. Windows-specific validation was not performed locally because the contributor's machine is enterprise-managed and blocks unsigned binaries; reviewers' Windows CI will exercise the change as part of the regular build matrix.

Repro steps that now pass:

1. Start Freelens with no synced kubeconfigs.
2. Open **Preferences → Kubernetes → Kubeconfig Syncs**.
3. Click **Sync Files and Folders** (or **Sync file(s)** / **Sync folder(s)** on Windows) and pick a kubeconfig file or folder.
4. The clusters appear in the catalog **immediately** — no restart, no extra clicks.
5. Remove a synced entry → the corresponding clusters disappear immediately.

## Notes for reviewers

- No platform-specific branching.
- No new IPC channels, no UI changes, no new dependencies.
- No DI registration changes — the modified descriptor and manager are the same injectables they were before.
- Lockfile and `package.json` are untouched.


yehuda-kahan:
Hi @robertobandini, I'm reaching out because I saw you were able to reproduce this on Windows back in [#551](https://github.com/freelensapp/freelens/issues/551#issuecomment-3575197729).

Can you please have a look at this fix proposal?

I was able to repro the bug and verify the fix end-to-end on Linux/WSL using a locally-built AppImage, but I couldn't validate it on Windows myself - my work machine is enterprise-managed and Defender blocks unsigned binaries from running, so I can't install a test build locally. The fix lives in a shared cross-platform code, so I'm fairly confident it behaves the same on Windows, but a real-Windows confirmation from someone who has already reproduced the bug would be hugely valuable before merge. So if you think it's worth it or needs this effort at all, and you can do it, I'd appreciate it as well.

Thank you, and happy to iterate on the patch if anything looks off.

dex4er:
@claude Check for regressions

Cenness:
Tried build from this action - https://github.com/yehuda-kahan/freelens/actions/runs/27298514731

Sync isn't happening on Windows. Both adding and removing a folder do nothing without restart.

dex4er:
@claude redo review for manager.ts using new AGENTS.md rules

dex4er:
@claude apply fixes and push to PR branch

dex4er:
@claude redo fix, push to new branch with `claude/` prefix, create a new PR for new branch then close this one.

yehuda-kahan:
> Tried build from this action - https://github.com/yehuda-kahan/freelens/actions/runs/27298514731
> 
> Sync isn't happening on Windows. Both adding and removing a folder do nothing without restart.

Hi, the mentioned build was generated to reproduce the issue before applying the fix. I couldn't test it on my PC because it's an enterprise PC, so I didn't generate a new one that contains the fix changes.

PR #1942, opened 2026-06-11
Display schedulerName in pod details

dwo:
**Description of changes:**

- Display the Kubernetes Pod `spec.schedulerName` field in the pod details panel as a `Scheduler` row, placed directly after the `Node` row (both are scheduling-related).
- Makes the scheduler that placed a pod visible at a glance (e.g. `default-scheduler` vs `kai-scheduler`) without inspecting raw YAML.
- Uses the existing `hidden={!schedulerName}` `DrawerItem` convention, consistent with the adjacent `Node` and `Runtime Class` rows. In practice `schedulerName` is API-defaulted to `default-scheduler`, so the row is shown on essentially every pod.
- No type or model changes were needed — `schedulerName?: string` already exists on `PodSpec`.

### DIFF
--- a/packages/core/src/renderer/components/workloads-pods/pod-details.tsx
+++ b/packages/core/src/renderer/components/workloads-pods/pod-details.tsx
@@ -71,11 +71,11 @@ class NonInjectedPodDetails extends React.Component<PodDetailsProps & Dependenci
         <DrawerItem name="Status">
           <span className={cssNames("status", kebabCase(pod.getStatusMessage()))}>{pod.getStatusMessage()}</span>
         </DrawerItem>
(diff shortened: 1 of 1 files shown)


dwo:
Would be open to hiding this when the value is `default-scheduler` if that is preferred :+1:

dex4er:
@dwo yes, I think it should be hidden for the default scheduler and I moved it before node as I think logically scheduler picks node, so the information should go in the ordering [who scheduled] -> [where scheduled]

dwo:
Awesome, thank you 🎉

dex4er:
Thanks!

PR #1949, opened 2026-06-12
Prune unused native binaries to prevent warnings about Intel-based apps in MacOS

dex4er:
Redo of #1923 as Claude massacred the previous PR


dex4er:
@claude provide the fix that implements checking if path exists more correct way for node

dex4er:
@claude provide patch in `diff -u` format

dex4er:
@claude review after changes

PR #1989, opened 2026-06-18
Remove the hotbar item limit

### LINKED ISSUE #848: Ability to add more clusters to hotbar

Hello, thank you for adding CPU and Memory columns.
Can you  make possibility to add more clusters to hotbar? I have plenty of them :)

Image

### Why is this needed?

It useful for those, who has many k8s clusters

1spyral:
Fixes #848

**Description of changes:**

- Remove the 12-item hotbar limit and empty cell placeholders.
- Allow hotbars to grow and scroll with any number of pinned items.
- Preserve all pinned items when loading legacy hotbar data.
- Remove fixed cells, which were originally intended for keyboard slot shortcuts. Shortcuts could still target the first 10 or 12 items without limiting the total number of pinned items.
- Update tests and snapshots for the new hotbar layout.


dex4er:
@1spyral what do you mean "remove empty cell placeholders"?

Because it sounds like the exact feature I use now: I put some clusters at the bottom of the menu then I really **need** empty placeholders.

1spyral:
@dex4er I removed them because a comment in the code said they were meant for keyboard shortcuts, which weren't implemented yet. I just thought of an alternate design: how about having a + cell to add more of these placeholder cells? And they can be removed on right click. Default # of placeholder cells will still be 12

dex4er:
@1spyral adding more placeholders with [+] is much better idea. If we would like to implement it, let's create it as a separate PR and close this one as invalid.

robertobandini:
Hi @1spyral thank you about your interest in Freelens and about the suggestions.
You are welcome to contribute with the new PR, thanks!

PR #2011, opened 2026-06-20
Disable release-asset upload in Create SBOM step

dex4er:
The `Create SBOM` step (`anchore/sbom-action@v0`) defaults to `upload-release-assets: true`, attaching the un-normalized lowercase SBOM directly to the release before normalization. The normalized copy is then attached again by `Upload files`, and on a re-run the stale asset triggers a `ReleaseAsset already_exists` error. Setting `upload-release-assets: false` makes anchore only produce the file.

**Description of changes:**

- Set `upload-release-assets: false` on the `anchore/sbom-action@v0` step in `.github/workflows/release.yaml`


dex4er:
@copilot move proposed fix to .github/workflows and push to this PR

PR #2016, opened 2026-06-22
Fix extension installs when user-data lacks package.json

### LINKED ISSUE #2015: Freelens extension installs fail when ~/.config/Freelens has no package.json
## Bug

On a clean user-data directory, Freelens tries to run pnpm in `~/.config/Freelens`, but that directory may not contain a `package.json`. pnpm then fails with `ENOENT: no such file or directory, stat "/home/ok/.config/Freelens/package.json"`, which breaks extension install/uninstall flows.

It happens during rpm package installation with prev version existance, also from rpm.

## Proposed fix

Bootstrap the pnpm root for the extension store before invoking pnpm, so the user-data directory always has a valid `package.json` for extension management. Add a regression test for the installer bootstrap path.

## Related PR

- https://github.com/freelensapp/freelens/pull/2014

kappa8219:
Bootstrap the pnpm root in ~/.config/Freelens before invoking extension installs, so a clean user-data directory no longer fails with ENOENT. Adds a regression test for the installer bootstrap path. Closes #2015


dex4er:
@claude Check this PR for regressions

dex4er:
@claude This PR is fixed but fix error message for failed writeJsonSync in a separate PR.

PR #2019, opened 2026-06-25
Use a light background for pod logs in the Light theme

s1ckoleg:
## What

In the Light theme, the pod logs panel still rendered dark (dark background, white text), which hurts readability. This makes the Light theme's pod logs light so the logs follow the selected theme. The Dark theme is unchanged.

| Theme | Before | After |
| --- | --- | --- |
| Light | dark logs (`#24292e` bg / white text) | white logs (`#ffffff` bg / `#2d2d2d` text) |
| Dark | black logs | black logs (unchanged) |

## Why

The logs panel is already theme-driven — `list.scss` reads the CSS variables `--logsBackground`, `--logsForeground`, and `--logRowHoverBackground`. The Light theme definition simply hardcoded *dark* values for those three colors (identical to the Dark theme). These values were inherited verbatim from the original OpenLens fork import and were never adjusted for the Light theme (`#24292e` is GitHub's dark code-block color).

Upstream Lens already addressed the same request — lensapp/lens#8137 "Add Light Theme for Pod Logs" — and shipped a light logs theme. This change brings Freelens in line with that behavior.

## Change

A single file, `packages/core/src/renderer/themes/lens-light.injectable.ts`:

- `logsBackground`: `#24292e` -> `#ffffff` (matches the Light theme's terminal background and content panes)
- `logsForeground`: `#ffffff` -> `#2d2d2d` (matches the Light terminal foreground; ~13.6:1 contrast on white)
- `logRowHoverBackground`: `#35373a` -> `#f0f2f5` (reuses the existing `sidebarItemHoverBackground` token)

All three reuse color values already present in the Light theme, so the logs stay visually consistent with the terminal and the rest of the light UI.

## Known limitation

ANSI-colored log output is rendered via `ansi_up`, which uses its own fixed color palette (no public override in v5). Uncolored log text falls through to `logsForeground` and reads fine on white. Logs that explicitly print white/bright-white foreground (rare) may be low-contrast on the white background. This matches how upstream shipped lensapp/lens#8137.

## Testing

- Built locally (`pnpm build && pnpm build:app:dir`) and ran the app.
- Opened a Pod's Logs and switched Preferences > Appearance > Theme between Light and Dark.
- Light: white background, dark text, light row hover. Dark: unchanged (black background, white text).
- `pnpm biome check` passes on the changed file; no unit tests or snapshots reference these theme color keys.


dex4er:
I would postpone it until we'll have support for a fully customized user's theme. This is a change, and any change, even good, might be bad for someone.

s1ckoleg:
Hi there! Fair point, wasn't awared about this feature in development
Dark theme in pod's logs on light theme just didn't make sense to me so I've decided to propose this fix (stuggle to read text on dark theme).

dex4er:
@s1ckoleg I think you're right. Probably someone who prefers a dark background for logs can just switch to dark theme.
Let's merge it!

robertobandini:
Thank you very much @s1ckoleg!

Issue #2061, opened 2026-06-29
kubectl.exe is missing


dex4er:
### Describe the bug

After opening the terminal: `Get-Command kubectl` shows "kubectl is not recognized".

### Freelens version

1.10.2

### Operating system type and version

Windows

### Installation method

MSI, EXE

dex4er:
@claude [runs-on:windows] Install MSI package and check if it contains bundled kubectl.exe

dex4er:
@claude Why kubectl.exe is not in the PATH?

Is it a regression, ie. after we enabled signing the bundled binaries or it was missing in the path always? 

Is it the same problem on non-Windows versions too?

dex4er:
@claude So it looks like downloading of the kubectl stopped working and fallback to use bundled kubectl too. Analyze the problem: why we don't have downloaded kubectl, and there is no any error in the console about failed download.

On the MacOS:

in the started terminal I have:

```
$ which kubectl
/Applications/Freelens.app/Contents/Resources/arm64/kubectl
```

so it looks like kubectl **is** in the PATH! Anyway: 

```
$ echo $PATH
/Users/dex4er/Library/Application Support/Freelens/binaries/kubectl/1.35.6:/Applications/Freelens.app/Contents/Resources/arm64:/Users/dex4er/.local/bin:...
```

so version 1.35.6 is not downloaded anymore.

Check if it is regression after https://github.com/freelensapp/freelens/pull/1825

dex4er:
@claude

The main log for MacOS now:

```
info:    ┏ Downloading kubectl 1.35.6 from https://dl.k8s.io/release/v1.35.6/bin/darwin/arm64/kubectl to /Users/dex4er/Library/Application Support/Freelens/binaries/kubectl/1.35.6/kubectl +379ms
```

and now the binary is available indeed. However, I can't reproduce the situation from the previous comment, when the kubectl has not been downloaded and then fallback (bundled version) has been used.

The main log for Windows:

```
info:    ┏ Downloading kubectl 1.35.6 from https://dl.k8s.io/release/v1.35.6/bin/windows/arm64/kubectl.exe to C:\Users\dex4er\AppData\Roaming\Freelens\binaries\kubectl\1.35.6\kubectl.exe +111ms
error:   ┏ Local kubectl failed to run properly (Error: spawn EBUSY), unlinking +201ms
error:   ┗ [1] { timestamp: '29/06/2026 19:36:17' }
```

Explain 2 issues:

[...]

PR #2062, opened 2026-06-29
Add bundled binaries directory to PowerShell and fish terminal PATH

dex4er:
Fixes part of #2061.

The integrated terminal only put the bundled `resources/<arch>` directory (which contains the kubectl shipped with the app) on PATH through the bash and zsh init scripts. PowerShell (the Windows default) and fish were launched with only the downloaded-binaries and user-data directories on PATH, so when the downloaded kubectl directory was empty there was no bundled fallback and `kubectl` was "not recognized".

This appends `baseBundledBinariesDirectory` to the PowerShell and fish PATH so the bundled kubectl resolves the same way it already does for bash/zsh.


dex4er:
@claude does it work on MacOS with Powershell too?

PR #2063, opened 2026-06-29
Keep kubectl binary when self-check hits a transient EBUSY error

dex4er:
Fixes part of #2061.

After downloading kubectl, `ensureKubectl()` immediately runs `kubectl version --client` to validate it. On Windows the freshly written executable is frequently locked for a short time by antivirus real-time scanning, so the probe fails with `spawn EBUSY`. `checkBinary()` treated any exec failure as a corrupt binary and unlinked it, leaving an empty version directory on PATH and "kubectl is not recognized".

Transient errors (EBUSY/ETXTBSY/EAGAIN/EPERM/EACCES) are now treated as non-fatal: the binary is kept and verifies on a later run. `checkBinary()` also now unlinks the `path` it was given rather than `this.path`.


dex4er:
@claude fix regression

./src/main/kubectl/kubectl.ts:42:61
  TS2345: Argument of type 'string | number' is not assignable to parameter of type 'string'.

etc

Issue #2102, opened 2026-07-08
Plan for Freelens v2


dex4er:
@claude [model:fable] 

Create the plan to change the bundler from webpack to vite with ESM-first modules, and jest to vitest.

Check https://github.com/freelensapp/freelens/issues/1718 for the context and previous analysis.

The development will be continued in the new "v2" branch that breaks compatibility with the previous versions.

We can spare in the 1st phase integration tests, unit tests, knip check, and possible additional GitHub Action tests.

The MVP will be the application that is compiled only from Typescript. If we'll need to reduce the number of modules: it'll be possible. Technically it is even possible to remove all modules and use the single workspace as far as internally it'll be possible to test current modules separately and they'll be somehow isolated to reduce internal dependencies but even this is not the strict requirement.

We'll use Typescript 7 that is superfast for bigger projects, but initially we can take the version that is currently available for Vite+Electron.

We can use any Electron version that supports ESM: it can be current version or any newer.

We can break compability for extensions however: we still will need to have this feature. New extension API also can use ESM modules or any other modules that will be supported by Electron. Any other changes to extension API will be tested with https://github.com/freelensapp/freelens-example-extension project then later we will prepare the instruction how to adapt the extension projects to the new Extension API.

In this iteration we'll make technical decisions.

dex4er:
@claude [model:fable] Freelens v2 is scaffolded. Go further with implementation.

dex4er:
@claude Add to AGENTS.md that we need to push after every commit to allow resuming after timeouted session. Total timeout is now 60m.

dex4er:
@claude [model:fable] Continue with the plan

dex4er:
@claude [model:fable] PR merged, go next

dex4er:
@claude [model:fable] go to next step. If any tests or checks are ready to enable, we can do it too.

dex4er:
@claude [model:fable] go to next step.

dex4er:
@claude [model:fable] go to next step.

dex4er:
@claude [model:fable] PR merged, go to next phase

dex4er:
@claude [model:fable] PR merged, go to next phase

We'll check the real extension later in a separate repo.

dex4er:
@claude [model:fable] PR merged, go to next phase

dex4er:
@claude [model:fable] PR merged, integration tests enabled, go to next phase

dex4er:
[...]

PR #2112, opened 2026-07-08
Build the app with electron-vite on the v2 branch

dex4er:
Continues #2102 on top of the v2 scaffolds (#2104–#2108).

Makes the electron-vite build real: `pnpm build` in freelens/ now produces an ESM static/build/main.js and a Vite renderer bundle from the TypeScript sources of all workspace packages in a single transpilation shot. Webpack stays available as build:webpack until Phase 5 removes it.

- Completes Phase 2: @freelensapp/core exports point at src, 37 package entries fixed (root index.ts), transpilation shims consumed from source
- electron.vite.config.ts completed: renderer node externals, SCSS modules, SVG ?raw, Monaco module workers, fonts as Vite assets, resource templates via import.meta.glob
- Extension loader works from ESM bundles via createRequire/require(esm) as verified in #1718

Not yet covered: dev-mode (electron-vite dev) integration with the lens proxy, runtime verification on a real Electron run, packaged app smoke test.

| Model: `claude-fable-5`


dex4er:
@claude add exception for licenses that caused failing https://github.com/freelensapp/freelens/actions/runs/28980994162/job/85999480352?pr=2112

PR #2119, opened 2026-07-10
Fix RoleBinding and ClusterRoleBinding create dialogs

giggsoff:
## Summary

Fixes two bugs, each present in both the **RoleBinding** and **ClusterRoleBinding** add/edit dialogs (`user-management/*/dialog/view.tsx`).

### 1. "Create" does nothing when creating a new binding

`createBindings` guarded on the dialog state's existing binding object (`roleBinding` / `clusterRoleBinding`), which is only set when **editing**. When creating a brand-new binding that value is `undefined`, so the guard returned early and `store.create` — and the dialog close — never ran. The Create button silently did nothing and the dialog stayed open.

Fix: only require the existing binding for the update path, and branch on it directly (instead of `isEditing`) so the create path runs when there is no existing binding. The `.create()` branch was effectively dead code before.

### 2. Selected service accounts are not shown

The Service Accounts multi-`<Select>` passed no `value` prop. `Select` is controlled and renders its selection solely from `value`, so chosen accounts were stored in the backing observable set but never displayed as chips — the selection looked lost.

Fix: pass `value={Array.from(this.selectedAccounts)}`, and drop the dead `isSelected` option flag the `Select` wrapper ignores.

## Changes

- `packages/core/src/renderer/components/user-management/role-bindings/dialog/view.tsx`
- `packages/core/src/renderer/components/user-management/cluster-role-bindings/dialog/view.tsx`
- Regression tests for the create flow in both dialogs' `__tests__/dialog.test.tsx` (assert `store.create` is called and `updateSubjects` is not).

Four commits, one per fix, for a bisectable history.

## Testing

- `pnpm test:unit` passes for both dialog suites. The new create-flow tests **fail against the pre-fix source** (`store.create` called 0 times) and pass with the fix.
- Verified end-to-end against a real cluster: creating a RoleBinding with User and ServiceAccount subjects now succeeds and closes the dialog; selected service accounts persist as chips in both dialogs.
- `pnpm biome check` clean on the changed files.

Generated with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/packages/core/src/renderer/components/user-management/cluster-role-bindings/__tests__/dialog.test.tsx
+++ b/packages/core/src/renderer/components/user-management/cluster-role-bindings/__tests__/dialog.test.tsx
@@ -5,6 +5,7 @@
  */
 
 import { ClusterRole } from "@freelensapp/kube-object";
+import { waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import React from "react";
 import directoryForKubeConfigsInjectable from "../../../../../common/app-paths/directory-for-kube-configs/directory-for-kube-configs.injectable";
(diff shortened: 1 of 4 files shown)


robertobandini:
Hi @giggsoff! Sorry for the long silence on this one, and thank you for the fix!

I have verified it end to end on the v2 main branch with a real cluster: without your change the Create button issues no request at all and the selected service accounts never show up, with your change both bindings are created with the right subjects and the dialogs close. Nice catch, this has been there since the original fork!

Main has moved to v2 and the unit tests now run on Vitest, so the four new mocks need `jest.fn` / `jest.Mock` replaced with `vi.fn` / `Mock` from vitest. The source changes rebase cleanly. We already have the rebase and the port ready locally: would you like to do it yourself, or are you fine with us pushing the rebased branch to your fork? Either way works for us!

One heads-up unrelated to your change: on current main, pages with many watches can leave requests queued for a while, so the dialog may look stuck after Create on a busy page. That is a separate v2 regression tracked in #2444, not this PR.

Small follow-up we noticed: pressing Enter in the Users field submits the wizard, so the binding gets created right away with the users typed so far. Same as edit mode today, worth a separate small fix.

@dex4er this is a small dialog-only fix with regression tests, orthogonal to the v2 work. Would you be fine with merging it on main for 2.0.0?

giggsoff:
> Hi @giggsoff! Sorry for the long silence on this one, and thank you for the fix!
> 
> I have verified it end to end on the v2 main branch with a real cluster: without your change the Create button issues no request at all and the selected service accounts never show up, with your change both bindings are created with the right subjects and the dialogs close. Nice catch, this has been there since the original fork!
> 
> Main has moved to v2 and the unit tests now run on Vitest, so the four new mocks need `jest.fn` / `jest.Mock` replaced with `vi.fn` / `Mock` from vitest. The source changes rebase cleanly. We already have the rebase and the port ready locally: would you like to do it yourself, or are you fine with us pushing the rebased branch to your fork? Either way works for us!
> 
> One heads-up unrelated to your change: on current main, pages with many watches can leave requests queued for a while, so the dialog may look stuck after Create on a busy page. That is a separate v2 regression tracked in #2444, not this PR.
> 
> Small follow-up we noticed: pressing Enter in the Users field submits the wizard, so the binding gets created right away with the users typed so far. Same as edit mode today, worth a separate small fix.
> 
> @dex4er this is a small dialog-only fix with regression tests, orthogonal to the v2 work. Would you be fine with merging it on main for 2.0.0?

Hi @robertobandini! Thanks for verifying it! I've rebased onto current main and ported the new mocks to Vitest (vi.fn / Mock), folded into the commits that introduced them so each commit stays green.

robertobandini:
Hi @giggsoff! Merged, thank you for the fix and for taking care of the rebase and the Vitest port yourself, everything was green on the first run.

The fix is on main and will ship with Freelens v2, which we are finishing right now. Thanks again for the contribution!

PR #2120, opened 2026-07-10
Build a publishable @freelensapp/extensions with a self-contained d.ts rollup

dex4er:
Continues #2102 on the v2 branch (Phase 4 of docs/v2-plan.md, the "fiddliest deliverable").

- The published @freelensapp/extensions now builds from three steps: whole-graph tsc --emitDeclarationOnly driven by a generated tsconfig whose paths map every workspace package entry to its TypeScript source; a rollup-plugin-dts pass that inlines all @freelensapp/* types into one self-contained dist/extension-api.d.ts; and a transpile of the runtime shim to a four-line ESM dist/extension-api.js over globalThis.FreelensExtensionApi.
- The workspace keeps consuming the TypeScript source entry; published entry points switch to dist/ via publishConfig and prepack runs the dist build. @freelensapp/core moves to devDependencies.
- Declaration emit surfaced one real bug (ws Server imported as a value where the ESM types export it type-only) and 14 TS2742 annotation gaps, fixed in separate commits.
- The first commit syncs .github/workflows/claude.yaml with main (b8b7ae8a): without it the app token cannot create any v2-based branch.

Verified: emit and bundle run clean; a strict-mode scratch consumer type-checks against the bundled d.ts. Not yet covered: the published dependency list for external type imports, the example-extension port, and a full build/test re-run (annotations are type-only).

| Model: `claude-fable-5`


dex4er:
@claude fix pnpm lock problems

dex4er:
@claude fix license check for rollup-plugin-dts package as it looks like a false alarm.

dex4er:
@claude Improve the workflow npm-licenses.yaml so we could have allowed_licenses in a separate dotfile in topdir instead of hardcoded in github workflow then further changes will be easier.

PR #2135, opened 2026-07-11
Keep test isolation and run the core Vitest project on the threads pool

dex4er:
Continues #2102 on the v2 branch and closes the last open Phase 7 item of docs/v2-plan.md, the test-run cost re-analysis.

- Records the maintainer decision: isolate: false is rejected because isolated tests have their own value. Prebundling or externalizing the @freelensapp/* packages for the test run is rejected for the same reason: modules evaluated once per worker instead of once per test file are shared state across test files. CI sharding of the core project is the steady-state approach.
- The core Vitest project runs on the threads pool: measured ~8% faster than the default forks pool on the import-dominated suite, with per-file isolation kept. Full suite green (2364 passed / 27 skipped) at 2.9 GB peak memory. vmThreads measured ~26% faster per shard but crashed the full suite at 13.7 GB peak memory and is rejected; the numbers are documented in the config and the plan.

Verified: full core suite green on threads; solution-wide type-check green; biome and trunk clean.

| Model: `claude-fable-5`

### DIFF
--- a/docs/v2-plan.md
+++ b/docs/v2-plan.md
@@ -190,11 +190,15 @@ Replace Jest 29 across the repository with Vitest 4:
 - Test-run cost: because every `packages/core` test file re-evaluates the
   eagerly-globbed injectable graph from TypeScript source, the core project
   dominates wall-clock (~57% of aggregate worker time is module `import`).
-  The interim mitigation is CI sharding of the core project (`vitest run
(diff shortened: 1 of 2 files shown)


dex4er:
@claude check of the tests results meet expectations and if we can merge this PR

PR #2137, opened 2026-07-11
Align TypeScript versions across packages via pnpm catalogs

### LINKED ISSUE #2132: Typescript v7 for Freelens v2
Currently we have different versions of Typescript used in project: 5.9.3 and 7.0.2. Actually Vite uses esbuild which doesn't use Typescript, so it is like 3rd flavour.

dex4er:
Aligns the per-package `typescript` devDependency through pnpm named catalogs so the version is declared once instead of being repeated in each package.json.

Two catalogs are kept intentionally: `ts7` (~7.0.2) for the repository-wide type-check and CLI emit, and `ts5` (^5.9.3) for tooling that imports the classic TypeScript compiler API (rollup-plugin-dts, typescript-plugin-css-modules, ts-node, ts-essentials, vitest-mock-extended). TypeScript 7 ships only the native tsc CLI and does not expose that API yet.

Resolves #2132.

### DIFF
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -147,7 +147,7 @@
     "electron-vite": "5.0.0",
     "playwright": "^1.61.1",
     "sass-embedded": "1.100.0",
-    "typescript": "catalog:ts5",
+    "typescript": "catalog:",
     "typescript-plugin-css-modules": "^5.2.0",
     "vite": "7.3.6",
     "vitest": "^4.1.10"
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -147,7 +147,7 @@
     "electron-vite": "5.0.0",
     "playwright": "^1.61.1",
     "sass-embedded": "1.100.0",
-    "typescript": "catalog:",
+    "typescript": "^6.0.3",
     "typescript-plugin-css-modules": "^5.2.0",
     "vite": "7.3.6",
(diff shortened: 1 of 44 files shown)


dex4er:
@claude Verify if we can upgrade rollup and other Vite components as Typescript 5 seems to be even more obsoleted today.

dex4er:
@claude Check if we can go to Typescript 6 instead, possibly as unified version.

dex4er:
@claude Let's go to option A as more stable and time below 30s seems to be still acceptable for now. Also supporting Renovate is important too and multiple version will be confusing.

dex4er:
@claude ok but clear the pnpm catalog and unnecessary comments.

dex4er:
@claude nah, we have renovate which scans versions then just don't use pnpm catalog

dex4er:
@claude now we have new regression after typescript upgrade. Review tsconfig options.

dex4er:
@claude Go with option A as we prefer single version and for TS7 we'll make yet another adaptation

dex4er:
@claude check if module resolution and TS target is consistent across all packages and eventually align settings.

dex4er:
@claude check if we still use deprecated node10 modes in tsconfig and replace settings with recommended for Electron 41.

PR #2143, opened 2026-07-12
Fix overlapping word-wrapped log rows by measuring row content height

s1ckoleg:
## Problem

With word wrap enabled, pod logs sometimes render as overlapping/stacked text. Long single-line JSON logs (especially with wide or non-ASCII glyphs) are the typical trigger.

Screenshot 2026-07-12 at 14 46 35

## Root cause

This is a regression from #2006. `LogList` estimates each wrapped row's height by character count and feeds it to react-window's `VariableSizeList`, which absolutely positions every row and pins its inline `height` to that estimate. The corrective measurement introduced in #2006 read `getBoundingClientRect().height` of the `.LogRow` element itself - but with the app's global `box-sizing: border-box`, that always returns the pinned estimate back. The measurement was a self-confirming no-op and could never correct an estimate error (the previous `scrollHeight - clientHeight` approach could at least detect undershoot).

When the estimate undershoots the real wrapped height, the row content overflows its unclipped box and paints over the rows below, and the error compounds down the list.

## Fix

- `list.tsx`: row content now lives in an inner wrapper div that lays out in normal flow; the measurement reads that wrapper's natural height plus the row's vertical padding. Both undershoots and overshoots self-correct on first render, without reintroducing the empty-space accumulation that #2006 originally fixed.
- `list.scss`: `overflow-y: clip` on `.LogRow.wordWrap`, so the single frame before a correction lands clips instead of painting over following rows. Nowrap mode is untouched and keeps horizontal scrolling.
- `list.tsx`: cached measured heights are invalidated when the timestamps toggle changes every line's length (they are keyed by row index and would otherwise go stale for off-screen rows).

## Testing

- New unit test asserting that the measured inner-content height plus the row's vertical padding is what reaches `VirtualList` as the row height.
- Full logs test suite passes (28/28), `biome check` and `trunk check` clean.
- Manual verification on the local build.


dex4er:
@claude Rebase againt the current main and see if the problem is not fixed yet

dex4er:
@claude create new PR after rebasing that will continue this PR

dex4er:
@s1ckoleg Thanks for your fix! I'll continue in the new PR as I can't use your own.

Issue #2145, opened 2026-07-12
Migrate legacy flexbox classes to tailwind


dex4er:
Previous issue #2138 detected problem with legacy flexbox CSS classes. Suggestion is to migrate them to the modern tailwind.

dex4er:
@claude [model:fable] Check v2 branch and prepare the plan to remove mixed classnames and use tailwind instead.

dex4er:
@claude Go to implementation. This time we are working on main branch.

dex4er:
@claude PR merged, go to another phase. Check #2147 for test results.

dex4er:
@claude Go to the next phase

dex4er:
@claude Go to the next phase

dex4er:
@claude next step

dex4er:
@claude Phase 1 is done, implement phase 2

dex4er:
@claude Prepare the PR for another bunch of commits for Phase 2

dex4er:
@claude PR merged, go to phase 3

dex4er:
@claude PR merged. Prepare the PR for another phase.

PR #2160, opened 2026-07-13
Migrate legacy flexbox to Tailwind: Input, Wizard, and the utility cascade-layer fix

dex4er:
Completes the `flexbox.scss` -> Tailwind migration for the remaining
first-party components and fixes a foundational cascade-layer bug that was
silently breaking Tailwind's margin/padding utilities across the whole app.

## Cascade-layer fix (the important one)

The global reset in `app.scss` (`*, *:before, *:after { margin: 0; padding: 0;
border: 0 }`) was **unlayered**. Unlayered rules win over any `@layer`
regardless of specificity, and Tailwind v4 emits utilities into
`@layer utilities`. So every margin/padding utility (`mb-5`, `ml-auto`,
`mt-3`, ...) was silently overridden everywhere -- only `gap`/flex utilities
survived, which is why earlier `gaps` -> `gap-*` batches looked fine while
standalone spacing utilities did nothing (e.g. the missing gap before "Synced
Items" under Kubeconfig Syncs).

Moving the reset into `@layer base` lets the later `utilities` layer win, while
component SCSS (still unlayered) keeps overriding both. This activates the
margin/padding utilities already present in ~21 core components.

## Migration completion

- Input: `flex gaps [align-center]` -> `flex [items-center] gap-2`;
  `box grow` -> `grow`; `box right` -> `margin-left: auto` in SCSS; removed the
  dead, never-interpolated `--flex-gap: $padding * 0.6667` no-op.
- Wizard buttons row: `flex gaps align-center` -> `flex items-center gap-2`.
- The guardrail baseline reached **0** -- the legacy flexbox migration for
  first-party code is complete -- so the ratchet guardrail
  (`scripts/check-legacy-flexbox.mjs` + baseline JSON + `lint:flexbox` scripts)
  is retired. New usage is kept out by convention (documented).

## Documentation and cleanup

- Document the cascade-layer rule in `docs/v2-styling.md`: Tailwind utilities
  live in `@layer utilities`, the app reset in `@layer base`, and unlayered
  component SCSS overrides utilities -- which is why the reset had to be layered
  for `m-*`/`p-*` to work at all. Replace the ratchet section with a
  migration-status note; add a FROZEN banner to `flexbox.scss`.
- Remove dead `--flex-gap` declarations left over from migrated components
  (`.AddClusters`, `.AddHelmRepoDialog`). `flexbox.scss` itself stays -- it is
  frozen for extension compatibility, not removed.

## Regression fixes surfaced during review

- **Add cluster**: the kubeconfig editor's fixed 600px height pushed the "Add
  cluster" button off-screen when a multi-line validation error appeared; height
  is now `clamp(300px, 50vh, 600px)`.
- **Hotbar spacing**: `flex column gaps` -> `flex flex-col gap-2` doubled the
  vertical spacing because `gap-2` is additive on top of each cell's own 12px
  margin (legacy `.gaps` overrode it). Dropped the redundant `gap-2`.
- **Hotbar hover shift**: `overflow: overlay` on hover made the scrollbar take
  layout width only while hovering, shifting the centered cells; now uses a
  constant-width scrollbar with `scrollbar-gutter: stable both-edges` and only
  toggles the thumb color on hover.
- **Port-forward buttons**: the "Forward..." buttons in Service and Pod details
  were ragged (they followed variable-length port text). Each port row is now a
  flex row that pushes the button to the right edge, so the buttons line up in a
  column.

Refs #2145

| Model: `claude-opus-4-8`


dex4er:
@claude Go to next iteration. Write what needs to be verified visually

PR #2162, opened 2026-07-13
Remove the legacy flexbox.scss utility layer

dex4er:
Removes the in-house `flexbox.scss` utility layer. The first-party migration to
Tailwind is complete (no core TSX, ui-components, or app code uses the legacy
classes), so the layer is no longer needed.

## Changes

- Delete `packages/core/src/renderer/components/flexbox.scss` and stop loading
  it in `app.scss`.
- Drop the now-dead `--flex-gap` defaults (`:root` in `app.scss`, `> *` in
  `wizard-layout.scss`) that only fed the old margin-based `.gaps`. The
  self-contained `--flex-gap` in `menu-actions.scss` stays — it is a local
  toolbar margin, not a `flexbox.scss` consumer.
- `docs/v2-styling.md` and `AGENTS.md`: `flexbox.scss` is no longer one of the
  styling systems; core uses Tailwind for layout. Removed the obsolete
  vocabulary/mapping/migration-status sections.
- `docs/v2-extension-migration.md`: new **Migrating off `flexbox.scss`** section
  with a legacy-class → plain-CSS mapping and a worked example for extension
  authors, plus a checklist item.

## Breaking change (extensions)

The host no longer provides the global flexbox classes (`flex`, `column`,
`gaps`, `box`, `grow`, `align-center`, …). Extensions that still use them must
add the equivalent CSS to their own stylesheet — see the new migration guide.

Audited local extensions — still using the classes (need migrating in their own
repos):

- **freelens-fluxcd-extension** — `box grow` ×3, `flex column` ×3, `column` ×1
- **freelens-gateway-api-extension** — `column` ×1

Clean: freelens-example, freelens-for-claude, freelens-ai (its `box` hits are
false positives, e.g. `text-input-select-box`).

Refs #2145

| Model: `claude-opus-4-8`


dex4er:
@claude [model:fable] Propose how to use tailwind with extensions.

dex4er:
@claude fix plan for extensions with the hint that they can use tailwind

PR #2163, opened 2026-07-14
feat: custom theme and accent color customization

elevasyncsolutions-jpg:
## Summary

Adds a custom theme/accent color feature that allows users to override the primary accent color of any Freelens theme from the Appearance settings page.

## Changes

- **New `customThemeColors` user preference**: Stores color overrides as a key-value map of `LensColorName` → hex color string, persisted across sessions via the existing `lens-user-store`
- **New `custom-theme-colors.injectable.ts`**: Provides the custom colors as a MobX computed value from user preferences state
- **Modified `active.injectable.ts`**: Merges custom color overrides on top of the selected base theme before applying it; if no custom colors are set, the base theme is returned unchanged
- **Modified theme settings UI**: Added an HTML5 color picker and Reset button in the Appearance section, allowing users to pick an accent color that automatically applies to `primary`, `blue`, `buttonPrimaryBackground`, `menuActiveBackground`, and `helmStableRepo`
- **Updated storage layer**: Loads/saves `customThemeColors` alongside other preferences

## Implementation Details

- Follows the existing DI pattern using `@ogre-tools/injectable`
- Custom colors are merged on top of the base theme in the active theme computed, so they work with both dark and light themes, as well as with system theme sync
- The `apply-lens-theme.injectable.ts` already sets CSS custom properties for all color keys, so overrides take effect automatically
- When the user resets the accent color, an empty object is stored, which the active theme resolution treats as no custom overrides

## Why not the previous approach (PR #1368)?

The previous PR was closed without merging, likely because it tried to do too much at once. This implementation keeps the change minimal:
- Only adds what is necessary (custom color storage + merge + UI)
- Reuses the existing theme infrastructure instead of creating a parallel theme system
- Does not modify the LensTheme interface or require theme authors to support customization

## Testing
- Select a theme, set a custom accent color, observe the UI updates
- Restart the app, verify the accent color persists
- Click Reset, verify the original theme colors are restored

### DIFF
--- a/packages/core/src/features/preferences/renderer/preference-items/application/theme/theme.tsx
+++ b/packages/core/src/features/preferences/renderer/preference-items/application/theme/theme.tsx
@@ -4,6 +4,7 @@
  * Licensed under MIT License. See LICENSE in root directory for more information.
  */
 
+import { action } from "mobx";
 import { withInjectables } from "@ogre-tools/injectable-react";
 import { observer } from "mobx-react";
 import React from "react";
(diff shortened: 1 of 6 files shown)


robertobandini:
Hi @elevasyncsolutions-jpg,

As explained in #1280 (see my comment there), the issue is still assigned to
gambhirsharma, so no further PRs are accepted, whether from humans or bots.
You can find our bounty policy here: #2293

For this reason I'm closing this PR without review.

In any case, everyone is welcome here! There are many ways to contribute to
Freelens and its extensions: open issues, documentation, the extensions
ecosystem, and of course other bounties when available. Take a look at
CONTRIBUTING.md and feel free to join us.

Thank you for your interest in Freelens!

PR #2182, opened 2026-07-14
Restore detail drawer toolbar icon spacing

### LINKED ISSUE #2181: Detail drawer toolbar icons lost their spacing (regression from flexbox.scss removal)
### Summary

In the kube-object detail drawer, the action icons on the right of the title bar (Attach to Pod, Shell, Logs, Edit, Delete, Force Delete) are packed edge-to-edge with **0 px between them**. Previously each icon had `margin-right: 8px`, so they were evenly spaced.

### Steps to reproduce

1. `pnpm dev` (or a packaged build), open a cluster, go to Workloads → Pods.
2. Click a pod to open its detail drawer.
3. Look at the icon row on the right of the green title bar.

### Observed vs expected

Measured over CDP in the running dev app (each icon is 21 px wide):

| icon | left–right |
|---|---|
| Attach to Pod | 1190–1211 |
| Shell | 1211–1232 |
| Logs | 1232–1253 |
| Edit | 1253–1274 |
| Delete | 1274–1295 |
| Force Delete | 1295–1316 |

Gap between consecutive icons = **0 px** (they touch). Only the trailing `close` (X) button keeps an 8 px gap, because that comes from the toolbar container's own `margin-right`, not from inter-item spacing.

Expected: 8 px between each icon (the previous behaviour).

### Root cause

Regression from #2162 (*Remove the legacy flexbox.scss utility layer*).

`.MenuActions.toolbar` never set its own inter-item spacing. It relied entirely on the `.flex.gaps` utility that `flexbox.scss` provided:

```scss
/* removed flexbox.scss */
.flex.gaps {
  $gap: var(--flex-gap, 1em);
  &:not(.column) > :not(:last-child) { margin-right: $gap; }
}
```

`menu-actions.tsx` still adds the now-dead class explicitly:

```tsx
// menu-actions.tsx:197
gaps: toolbar, // add spacing for .flex
```

and `menu-actions.scss` still declares `--flex-gap: var(--padding)` (= `var(--unit)` = 8 px), now an orphan. With `flexbox.scss` gone, the `gaps` class matches no rule, so the 8 px inter-icon margins disappeared. A codebase scan shows `menu-actions` is the only remaining consumer of `gaps` / `--flex-gap`, so this is the only regression of this kind.

### Proposed fix

Restore the spacing with a modern `gap` directly on the toolbar (the container is already `display: flex`), rather than reintroducing the removed `flexbox.scss` utilities (which `AGENTS.md` forbids):

```diff
--- a/packages/core/src/renderer/components/menu/menu-actions.scss
+++ b/packages/core/src/renderer/components/menu/menu-actions.scss
@@
   &.toolbar {
     --flex-gap: var(--padding);

     position: static;
     padding: 0;
     background: none;
     border: none;
     flex-direction: row;
+    gap: var(--flex-gap);
     margin-right: var(--flex-gap) !important;
```

`var(--flex-gap)` resolves to 8 px, exactly matching the previous `margin-right`. The now-dead `gaps` class (and its `// add spacing for .flex` comment) in `menu-actions.tsx:197` can be dropped as a follow-up cleanup, since nothing styles `.gaps` anymore.

### Notes

- Found while validating native dev mode (#2136) via a frame-aware CDP client.
- Worth double-checking whether #2162 dropped inter-item spacing anywhere else, though the scan above suggests `menu-actions` was the last `.gaps` consumer.

dex4er:
Restores the 8 px spacing between the kube-object detail drawer toolbar action icons, lost when the legacy flexbox.scss utility layer was removed in #2162.

Uses a modern `gap` on the toolbar container and drops the now-dead `gaps` class.

Fixes #2181

### DIFF
--- a/packages/core/src/features/catalog/__snapshots__/opening-entity-details.test.tsx.snap
+++ b/packages/core/src/features/catalog/__snapshots__/opening-entity-details.test.tsx.snap
@@ -4295,7 +4295,7 @@ exports[`opening catalog entity details panel > when navigated to the catalog > 
           </div>
         </div>
         <ul
-          class="Menu MenuActions flex CatalogEntityDrawerMenu toolbar gaps right bottom"
+          class="Menu MenuActions flex CatalogEntityDrawerMenu toolbar right bottom"
           id="menu-actions-for-catalog-entity-drawer-menu-some-entity-id"
         >
           <li
--- a/packages/core/src/features/cluster/kube-object-details/extension-api/__snapshots__/disable-kube-object-detail-items-when-cluster-is-not-relevant.test.tsx.snap
+++ b/packages/core/src/features/cluster/kube-object-details/extension-api/__snapshots__/disable-kube-object-detail-items-when-cluster-is-not-relevant.test.tsx.snap
@@ -37,7 +37,7 @@ exports[`disable kube object detail items when cluster is not relevant > given e
             </div>
           </div>
           <ul
-            class="Menu MenuActions flex KubeObjectMenu toolbar gaps right bottom"
+            class="Menu MenuActions flex KubeObjectMenu toolbar right bottom"
             data-testid="menu-actions-for-kube-object-menu-for-some-uid"
             id="menu-actions-for-kube-object-menu-for-some-uid"
           />
@@ -788,7 +788,7 @@ exports[`disable kube object detail items when cluster is not relevant > given e
             </div>
           </div>
           <ul
-            class="Menu MenuActions flex KubeObjectMenu toolbar gaps right bottom"
+            class="Menu MenuActions flex KubeObjectMenu toolbar right bottom"
             data-testid="menu-actions-for-kube-object-menu-for-some-uid"
             id="menu-actions-for-kube-object-menu-for-some-uid"
           />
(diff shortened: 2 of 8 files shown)


dex4er:
### Verified locally

Ran this branch with `pnpm dev` (Electron 41, Node 24) and drove the app through a frame-aware CDP client (Playwright MCP over the dev `--remoteDebuggingPort 9223`), reaching into the cross-origin cluster iframe (`iframe[name="orbstack"].contentFrame()`): Welcome → Browse Clusters → open cluster → Workloads → Pods → open a pod's detail drawer.

Measured the drawer header action-icon toolbar (`menu-actions-for-kube-object`, 6 icons: Attach to Pod, Shell, Logs, Edit, Delete, Force Delete):

| | before (`main`) | this branch |
|---|---|---|
| toolbar `gap` | `normal` (unset) | `8px` |
| gaps between consecutive icons | `[0, 0, 0, 0, 0]` px | `[8, 8, 8, 8, 8]` px |

The 8 px spacing is restored, matching the pre-#2162 `margin-right`. The `close` (X) button keeps its own 8 px separation as before. No new console errors attributable to this change (the one `Warning: Encountered two children with the same key ... k3s-Update` is a pre-existing, unrelated issue in the Managed Fields section).

dex4er:
@claude refresh snapshots

Issue #1333, opened 2025-11-09
Empty space in ingress view


Cenness:
### Describe the bug

If ingress has a lot of balancers or rules, freelens draws increasingly huge margin before next ingress, depending on its count.

### Reproduction steps

1. Create a few ingresses
2. Add more than two rules to one of them
3. Look at empty space after "N more"

### Expected behavior

No additional empty space per rule/balancer

### Screenshots

<img width="528" height="1004" alt="Image" src="https://github.com/user-attachments/assets/0fdaae16-aafc-42bb-bf1d-2b0e96ee5fc0" />

### Freelens version

1.6.1

### Operating system type and version

Windows 10

### Installation method

MSI

### Kubernetes cluster version used with Freelens

any

### Additional context

Related: https://github.com/freelensapp/freelens/issues/1029, https://github.com/freelensapp/freelens/pull/1146

PR #2195, opened 2026-07-15
Fix ingress row height when routes overflow

Strekol:
The Ingresses list virtualized each row's height using the full route count, but the cell only ever renders up to 1 route line (plus an overflow indicator). When an ingress had many rules, its row reserved far more height than was actually painted, pushing subsequent rows out of view.

Cap the reserved height to what showRoutes() actually renders by sharing a single MAX_VISIBLE_ROUTES constant and a helper that mirrors showRoutes()'s line count (rule lines shown, plus one line for the 'N more...' overflow indicator).


Cenness:
Thanks. Should close #1333

robertobandini:
@Strekol Hi and thank you very much about your contribute!

Strekol:
> @Strekol Hi and thank you very much about your contribute!

You're welcome :)

PR #2196, opened 2026-07-15
Make Wizard dialogs follow the dark theme

### LINKED ISSUE #2161: Wizard does not respect the dark theme (dark text on light background)
## Description

In the **dark theme**, the `Wizard` component does not follow the theme: it
renders with a light/white background and dark (or low-contrast) text. The
wizard's stylesheets carry hardcoded, non-theme colors that do not adapt to the
active theme, and some themed values appear to resolve to light in the wizard
context.

## Steps to reproduce

1. Use the dark theme (default).
2. Open any dialog that renders a `Wizard`, e.g.:
   - Port Forward dialog (Network → Services / Pods → Forward…)
   - Scale dialog (Deployment / StatefulSet / ReplicaSet)
   - Add namespace / Add secret / Add resource quota dialogs
   - Kubeconfig dialog, Helm release upgrade dialog
3. Observe the wizard body/stepper: light background with dark/low-contrast
   text instead of the dark-theme colors.

## Expected

The wizard should follow the active theme (dark background + light text in the
dark theme) like the rest of the dialogs, using the theme CSS custom properties
(`var(--…)`) rather than hardcoded colors.

## Likely causes / leads

Hardcoded, non-theme colors in the wizard stylesheets (do not adapt to the
theme):

- `packages/core/src/renderer/components/wizard/wizard.scss`
  - `> .header { color: white; }` (line ~33)
  - `.Stepper { color: silver; --stepper-color-point: black; }` (lines ~50-51)
- `packages/core/src/renderer/components/layout/wizard-layout.scss`
  - `.content-col > .error { background-color: pink; }` (line ~43) — this looks
    like a leftover debug placeholder and should be a theme error color.

The themed backgrounds it does use (`var(--dialogBackground)`,
`var(--dialogTextColor)`, `var(--contentColor)`) should be checked to confirm
they resolve to dark-theme values in the wizard/dialog context; if the wizard
renders in a portal outside the theme scope, the tokens may fall back to light.

## Notes

Noticed during the legacy-flexbox → Tailwind migration QA (#2145); this is a
separate pre-existing theming bug, unrelated to that migration.

dex4er:
Fixes #2161

**Description of changes:**

The `Wizard` component rendered with a light background and low-contrast text in the dark theme, because the dialog-specific theme tokens and some wizard stylesheets carried hardcoded, non-theme colors.

- **`lens-dark` theme**: give dark values to `dialogBackground` (`#262b2f`, matching `contentColor`) and `dialogFooterBackground` (`#2e3136`), previously `#ffffff` / `#f4f4f4`. These tokens are used **only** by the wizard, so this only affects the wizard body/footer; the light theme keeps its light values.
- **`wizard.scss` Stepper**: drop the hardcoded `color: silver` and `--stepper-color-point: black`; point the inactive step color at `var(--borderColor)` and let the Stepper inherit the themed text colors from `stepper.scss`.
- **`wizard-layout.scss`**: replace the leftover `background-color: pink` debug placeholder on `.error` with `var(--colorError)` + white text, matching the existing `InputTooltipError` convention.

The wizard header keeps white text on purpose: its background (`dialogHeaderBackground`) is a dark title bar in **both** themes, so white is the correct, contrasting color.

**Verification (Playwright, dev app, orbstack cluster):**

- Dark theme: Scale Deployment dialog and Create Namespace dialog now render dark (dark body/footer, readable text). 
- Light theme: dialogs unchanged (light body, dark title bar) - no regression.
- Note: the `Stepper` only renders when a `Wizard` has more than one step; no current dialog does, so that change is theme-correct but not visible in the running app.

_Screenshots added below._


dex4er:
wizard-namespace-dark
wizard-namespace-light
wizard-scale-dark

PR #2198, opened 2026-07-15
Upgrade Electron to v42

### LINKED ISSUE #2192: Upgrade Electron to v42
Probably we can go now to Electron v42.

Noticable breaking change might be:

* Behavior Changed: electron no longer downloads itself via postinstall script

and it might change our CI scripts.

dex4er:
Bumps Electron from `^41.10.0` to `^42.7.0` across the app and workspace packages, the extensions peer dependency range, and the electron-builder version pin.

Electron 42.7.0 bundles Node.js 24.18.0, which already matches the version pinned in .nvmrc, mise.toml and .trunk/trunk.yaml - no Node bump was required.

Fixes #2192

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,13 @@ download the right package for your system.
 
 ### macOS
 
-macOS 12 or later is required.
+macOS 13 or later is required.
 
 Download either the PKG (installer) or DMG (image) package from the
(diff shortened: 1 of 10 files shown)


dex4er:
@claude type check failed, so fix it

PR #2205, opened 2026-07-16
Append current timestamp to downloaded log filenames

shurkanTwo:
## Problem

Downloaded pod log files use a fixed name (`<pod>.log` for visible logs, `<container>.log` for all logs). Downloading logs twice for the same pod or container silently overwrites the previous file, which loses data when comparing logs over time.

## Change

- New helper `getTimestampedLogFilename(prefix)` builds `<prefix>-<ISO timestamp>.log`, with colons and dots replaced by dashes so the name is safe on all filesystems, e.g. `docker-exporter-2026-05-21T09-55-41-123Z.log`.
- `LogTabViewModel.downloadLogs` and the download-all-logs injectable use the helper instead of the fixed names.
- Unit test for the helper (fake timers) plus updated assertions in `download-logs.test.tsx`.

## Validation

- `pnpm biome check` on the changed files: clean.
- `vitest run` over `packages/core/src/renderer/components/dock/logs/`: 8 files, 29 tests passed.
- Repo `type-check` (generated tsconfig + `tsc`): no errors.

### DIFF
--- a/packages/core/src/features/pod-logs/download-logs.test.tsx
+++ b/packages/core/src/features/pod-logs/download-logs.test.tsx
@@ -177,7 +177,11 @@ describe("download logs options in logs dock tab", () => {
           });
 
           it("shows save dialog with proper attributes", () => {
-            expect(openSaveFileDialogMock).toHaveBeenCalledWith("dockerExporter.log", "some-logs", "text/plain");
+            expect(openSaveFileDialogMock).toHaveBeenCalledWith(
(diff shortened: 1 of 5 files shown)


robertobandini:
Hi @shurkanTwo thank you about your interest and your contributes!
Could be you have to wait until the job about Freelens v2 is finished.

shurkanTwo:
@robertobandini Is the job with Freelens v2 already done now?
What do you suggest how we should proceed with this PR?

I would really like to see log file names with their download timestamp.
It's increasingly annoying to manually rename log files each time I download them.

PR #2210, opened 2026-07-17
Inline await-lock source and drop CJS interop boilerplate

### LINKED ISSUE #2209: await-lock bundled
We have unnecessary configuration to handle await-lock module (additional wrapper, Vite configuration). All it might be removed when we'll just bundle original await-lock provided as Typescript.

dex4er:
Vendors the original await-lock v2.2.2 TypeScript source directly into `common/utils/await-lock.ts`, removing the runtime interop wrapper and the `bundledCjsPackages` entry in `electron.vite.config.ts`, and drops the `await-lock` dependency.

Fixes #2209.


dex4er:
Replaced by #2211

PR #2217, opened 2026-07-17
Replace grapheme-splitter with native Intl.Segmenter

### LINKED ISSUE #2170: Replace grapheme-splitter with native Intl.Segmenter
grapheme-splitter is deprecated in favor of the native `Intl.Segmenter`. Used in a single file: `packages/core/src/common/catalog/helpers.ts`.

dex4er:
Replaces the deprecated `grapheme-splitter` package with the native `Intl.Segmenter` API in `packages/core/src/common/catalog/helpers.ts`, and removes the dependency from both packages and the lockfile.

Fixes #2170

| Model: `claude-opus-4-8`

### DIFF
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -83,7 +83,6 @@
     "electron-window-state": "^5.0.3",
     "fs-extra": "^11.3.6",
     "glob-to-regexp": "^0.4.1",
-    "grapheme-splitter": "^1.0.4",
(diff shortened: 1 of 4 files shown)


dex4er:
@claude Check why we have difference in snapshots as theoretically using native grapheme split method should not bring any changes.

dex4er:
@claude put the proposed fix to this PR

PR #2247, opened 2026-07-20
Move electron from dependencies to devDependencies

### LINKED ISSUE #2246: Move electron dependencies to devDependencies
A few packages has

```
"electron": "^42.7.0"
```

still in dependencies rather than devDependencies in packages/technical-features/*

Also we should check if we can move more packages there from runtime to devtime.

dex4er:
Moves `electron` from `dependencies` to `devDependencies` in the three technical-feature packages that only need it at build/test time (electron is provided by the host app at runtime, matching `packages/core` and `freelens`):

- @freelensapp/application-for-electron-main
- @freelensapp/messaging-for-main
- @freelensapp/messaging-for-renderer

A broader scan of other runtime deps found the remaining build/test tooling in `dependencies` to be intentional (published types, test-support re-exports, build tools).

Closes #2246

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -167,6 +167,22 @@
       // only explicit react import here is type-only (ButtonHTMLAttributes),
       // so react looks unused even though it is a real runtime dependency.
       "ignoreDependencies": ["react"]
+    },
+
+    // electron is provided by the host app at runtime, so it is declared only
+    // as a devDependency; production code in these packages imports it, which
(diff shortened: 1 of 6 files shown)


dex4er:
@claude Now fix knip check

dex4er:
@claude Check the latest test result https://github.com/freelensapp/freelens/actions/runs/29729226868/job/88310730322?pr=2247 and see if we can tune knip.json for hints, but remember, we have single configuration for production and development build.

PR #2266, opened 2026-07-21
Replace react-router matchPath with an in-house port

dex4er:
Replaces react-router 5's `matchPath` -- the only pure-function react-router piece still used, across 6 call sites -- with a faithful in-house port in `@freelensapp/routing` over `path-to-regexp` v1, the engine react-router 5 bundles internally and the dialect Freelens' route schemas are authored in (/:param? optionals and inline /:param(regex) patterns). Adds unit tests over the real route patterns and the extension URL schema.

This is PR 2 of the follow-up sequence in the Phase 2 scoping document (docs/v2-routing-modernization.md).

Part of #2261.

| Model: `claude-opus-4-8`

### DIFF
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -94,7 +94,7 @@
     "node-fetch": "^3.3.2",
     "node-pty": "1.2.0-beta.14",
     "p-limit": "^7.3.0",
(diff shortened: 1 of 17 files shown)


dex4er:
@claude fix knip check

PR #2267, opened 2026-07-21
Vendor path-to-regexp v1 into @freelensapp/routing and drop the external dependency

dex4er:
Vendors a faithful TS/ESM port of `path-to-regexp` v1.9.0 into `packages/routing/src/vendor/path-to-regexp.ts` so the v1 dialect stays the routing contract and the code is owned in-repo. `match-path` uses the vendored engine, a purpose-named `compileRoutePath` wrapper is exported, the protocol-handler validates via it, and `buildUrl`'s `fillPath` is rewritten as a small regex substitution. `path-to-regexp` is removed as a direct dependency of `routing`, `core`, `utilities`, and `freelens`.

This is PR 3 of the Phase 2 routing-modernization sequence.

Part of #2261.

| Model: `claude-opus-4-8`

### DIFF
--- a/freelens/package.json
+++ b/freelens/package.json
@@ -94,7 +94,6 @@
     "node-fetch": "^3.3.2",
     "node-pty": "1.2.0-beta.14",
     "p-limit": "^7.3.0",
-    "path-to-regexp": "^1.9.0",
     "pnpm": "^11.15.0",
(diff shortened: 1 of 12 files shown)


dex4er:
@claude add vendored library to the generated licenses (pnpm build:resources:license)

PR #2269, opened 2026-07-21
Replace react-router Redirect / Switch / Route with in-house components

dex4er:
Adds in-house `<Route>` / `<Switch>` / `<Redirect>` to `@freelensapp/routing` over the in-house observable history and `matchPath`, reproducing the react-router v5 surface Freelens uses (exact-vs-prefix Route matching, first-match-wins Switch ordering, Redirect navigate-on-mount). Rewires the remaining react-router JSX call sites (`tab-layout.tsx`, the cluster-manager and cluster-frame redirect sites) and removes the now-unused `<Router>` HOC. This is step 4 of the Phase 2 routing-modernization sequence; the `react-router` dependencies and `toHistoryV4` shim are left for step 5.

Part of #2261.

| Model: `claude-opus-4-8`


dex4er:
@claude Check problem from Trunk Check

If react-router is vendorized then maybe we need to add it to `generate-license` licenses?

PR #2276, opened 2026-07-22
Add node and vitest/globals to tsconfig types

arnaud-dezandee:
**Description of changes:**

- Add an explicit `compilerOptions.types` array (`["node", "vitest/globals"]`) to the root `tsconfig.json` and the shared `packages/infrastructure/typescript/config/base.json`.
- Makes Vitest globals (`describe`, `it`, `expect`, etc.) and Node types available to the type checker without a per-file import, so test files can rely on ambient globals.


dex4er:
@arnaud-dezandee Good eye, thanks! However, currently our `globals: true` settings is temporary as the settings will change after transition to React 18 or 19.

I'm putting it on the list for post-transition tasks.

dex4er:
@arnaud-dezandee Actually you can give me a hint, if we should disable any globals, then use explicit imports from vitest or use it as a global. However, I would like to avoid adding anything other than nodejs globals to the default settings.

Even right now we have several conflicts between nodejs vs WHATWG types.

Anyway, I would like to postpone the discussion and decision after the main transition.

dex4er:
@claude should we add vitest globals to the separate tsconfig or use explicit imports? We want:

1. Not break type checks for tests and normal code
2. Editor should not show us false errors
3. Globals should not be too wide, as ie. vitest globals should not be applied to the normal code

dex4er:
@claude Rerun the recommendation after React transition is made and now we use React v19

PR #2303, opened 2026-07-23
Remove stale react-router leftovers from knip.jsonc

dex4er:
Removes dead `ignoreDependencies` entries for `react-router`, `react-router-dom`, their `@types/*`, and `mobx-observable-history` from `knip.jsonc`. These packages were removed from all manifests during the routing modernization, so the knip config entries were dead.

Closes #2175

Generated with [Claude Code](https://claude.ai/code)

PR #2305, opened 2026-07-23
Break the electron cache miss loop and retry Build Electron app

dex4er:
Implements the recommendations from [Claude's CI failure analysis on #2303](https://github.com/freelensapp/freelens/pull/2303#issuecomment-5062370209).

## Problem

The `integration-tests` workflow's `Use Electron cache` / `Use Electron Builder cache` steps use the combined `actions/cache` action, whose save only runs in a post-job step. That post step is skipped once a later step in the job fails.

When `Build Electron app` fails after the Electron / electron-builder binaries have already finished downloading (e.g. a transient HTTP 500 from the electron-builder binary mirror while packaging on Windows), the cache never gets written. The next run misses the cache again, re-downloads the same binaries, and can hit the same flaky upstream - a self-perpetuating miss.

## Fix

- **Break the cache loop:** split both the Electron and Electron-Builder caches into explicit `actions/cache/restore` (early, unchanged position) + `actions/cache/save` with `if: always()` (right after the `Build Electron app` step and its retry). This is the documented `actions/cache` pattern for saving a cache even when a later step in the job fails.
- **Retry the packaging step:** wrap `Build Electron app` in the same `continue-on-error: true` + conditional retry pattern already used elsewhere in this workflow (`Install pnpm dependencies`, `Build extra resources`, the per-OS integration test steps), to absorb a single transient failure like the HTTP 500 seen on #2303.

## Test plan

- [x] `yq e '.' .github/workflows/integration-tests.yaml` - YAML parses
- [x] `trunk check .github/workflows/integration-tests.yaml` - no issues
- [ ] CI run on this PR exercises the new restore/save/retry steps

Closes discussion on #2303.

| Model: `claude-sonnet-5`

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -53,7 +53,6 @@ jobs:
 
     env:
       DO_NOT_TRACK: "1"
-      ELECTRON_BUILDER_CACHE: ${{ github.workspace }}/.cache/electron-builder
       TURBOREPO_CACHE: ${{ github.workspace }}/.turbo
 
     steps:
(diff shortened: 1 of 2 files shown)


dex4er:
@claude analyze test run and check why Electron cache is not built on Windows

PR #2330, opened 2026-07-24
Move shared dependencies into the pnpm catalog

### LINKED ISSUE #2324: Expand pnpm catalog to shared dependencies
## What would you like to be added?

Move widely-shared dependencies into the pnpm `catalog` in `pnpm-workspace.yaml` (as already done for the React family), instead of declaring versions independently in each package's `package.json`. Keep single-package dependencies local. Do **not** push everything back to individual `package.json` files.

## Why is this needed?

Dependencies are heavily duplicated across the 46 workspace packages:

| Dependency | Declared in |
|---|---|
| `vitest` | 31 packages |
| `typescript` | 28 packages |
| `mobx` | 20 packages |
| `react` (already cataloged) | 18 packages |
| `@types/node` | 13 packages |
| `mobx-react` | 9 packages |
| `type-fest` | 8 packages |

- **Singletons.** `react` is already cataloged because multiple copies break hooks. `mobx` / `mobx-react` are the same class of problem — today they are held in line with the `mobx-react` override + `peerDependencyRules`. A catalog makes version alignment the default.
- **Upgrade ergonomics.** Bumping `vitest` currently means editing ~31 files; with a catalog it is a single line and a small PR (the exact pain seen when splitting the recent upgrade work into separate issues).
- **Consistency of direction.** The React family already validates the pattern; extending it is not a new paradigm.

## Scope — what to catalog

Singletons + pervasive tooling:

- `mobx`, `mobx-react`, `mobx-react-lite`
- `typescript`
- `vitest`, `@testing-library/*`, `jsdom`, `vitest-mock-extended`, `vitest-canvas-mock`
- `@types/node`, `type-fest`
- candidates: `chart.js`, `monaco-editor`, `immer`

Keep **single-package** dependencies in their own `package.json` (a catalog entry for a dep used once only adds indirection). If a package genuinely needs a divergent version, use a **named catalog** (`catalogs:`) rather than dropping it from the catalog.

## What a catalog does NOT replace (leave as-is)

- **Security / transitive overrides** — `form-data`, `tmp`, `brace-expansion` pin *transitive* versions; the catalog only affects direct deps in workspace packages, so these overrides stay.
- **`mobx-react` override** also forces transitive alignment (e.g. via `@ogre-tools`), so the catalog only helps the direct-dep side; keep the override + `peerDependencyRules`.

## Publish caveat

Published packages must have `catalog:` resolved to a concrete version at pack/publish time. pnpm does this automatically, and the repo **already publishes packages that use `catalog:` for react** (e.g. `keyboard-shortcuts`, `metrics`, `animate`), so the release pipeline already handles the rewrite. Risk is effectively zero as long as publishing goes through `pnpm pack`/`publish`.

## Renovate impact

The current `.renovaterc.json5` needs little to no change:

- **Works out of the box.** Renovate's pnpm manager parses `catalog`/`catalogs` and opens PRs editing `pnpm-workspace.yaml`; React is already managed this way.
- **Name/datasource rules carry over automatically** (they match by dep name/datasource, not by file), so these keep working after a dep moves into the catalog:
  - `@types/node` major/minor disabled (stays on 24.x)
  - `electron` major disabled
  - `node-pty` `deb` versioning
  - `ogre-tools monorepo` group
  - `matchDatasources: npm → minimumReleaseAge 25h`
- **One behavioral change to decide on:** `depType`-scoped rules do not see catalog entries. The `matchDepTypes: ["peerDependencies"] → rangeStrategy: widen` rule will **not** apply to a cataloged peer dependency (its version lives in the catalog as a single shared range, bumped by the global `bump` strategy). If widening peer ranges matters, keep those peers out of the catalog or put them in a named catalog with a dedicated rule.
- **Migrate each dependency atomically** (all consumers → `catalog:` in one change). A half-migrated dep (some packages cataloged, some with explicit versions) makes Renovate open two update streams for the same package.
- Optional cosmetic: a `packageRule` with `matchFileNames: ["pnpm-workspace.yaml"]` or a "catalog updates" group/label to collect catalog bumps.

## Tasks

- [ ] Add the shared deps above to the `catalog` in `pnpm-workspace.yaml` (consider named catalogs for any divergent versions).
- [ ] Replace their versions in each consuming `package.json` with `catalog:` (atomically per dependency).
- [ ] Decide on the peer-dependency widening question (pnpm-workspace catalog vs. named catalog vs. keep local).
- [ ] `pnpm install`; verify the lockfile is consistent and no unintended version changes occur.
- [ ] `pnpm type:check` / `pnpm biome check` / `pnpm build` pass; run the unit suite.
- [ ] Verify a published package still resolves `catalog:` to a concrete version on `pnpm pack` (dry-run pack of one SDK package).

dex4er:
Migrates widely-shared dependency versions out of individual package.json files and into the default pnpm catalog in pnpm-workspace.yaml.

All consumers already used identical versions, so a single default catalog covers them and the install produces no dependency-resolution changes. Publishing verified via pnpm pack dry-run.

Fixes #2324.

| Model: `claude-opus-4-8`


dex4er:
@claude Check if we can move even more, ie. everything except overrides.

dex4er:
@claude ok let's go

dex4er:
@claude move a few leftovers from devDependencies from freelens/package.json and others too as it'll be easier to manage everything from one place.

dex4er:
@claude Align devDependencies from topdir with the rest of the project. Either use "^" or pin the version for pnpm and turbo everywhere.

Issue #1051, opened 2025-08-05
Natural Text Editing doesn't work in terminal with ZSH


kdeyko:
### Describe the bug

The "Option (Alt) + arrow" navigation (aka "Natural Text Editing") doesn't work in Freelens' terminal with ZSH: instead of moving the cursor between words, it just prints `;5D` (Option + left) or `;5C` (Option + right).
This works in OpenLens, iTerm2 and macOS Terminal.
Surprisingly, it works in **bash** in Freelens.
For context, I have this zsh magic in my `.zshrc`:
```bash
# Fix CMD+Backspace behavior
# https://stackoverflow.com/questions/6205157/how-to-set-keyboard-shortcuts-to-jump-to-beginning-end-of-line/29403520#29403520
bindkey "^U" backward-kill-line

# https://unix.stackexchange.com/a/626504
# Use bash-like word definitions for navigation and operations
autoload -Uz select-word-style
select-word-style bash
# Use C-w to kill back to the previous space
zle -N backward-kill-space-word backward-kill-word-match
zstyle :zle:backward-kill-space-word word-style space
bindkey '^W' backward-kill-space-word
```
---
Additionally, CMD+Backspace should delete the entire line of text, while now it just deletes a single character.
This work in iTerm2.

### Reproduction steps

1. Open terminal in Freelens, using macOS ZSH as shell
2. Type `foo bar baz`; at this moment, your cursor will be right after `baz`
3. Press "Option+left" or "CMD+Backspace"

### Expected behavior

Option+Arrow: Cursor should move between words.
CMD+Backspace: The complete line should be deleted.

### Freelens version

1.5.2

### Operating system type and version

macOS Sequoia 15.6

### Installation method

brew

robertobandini:
Hi @kdeyko thank you about reporting this and about your interest about Freelens!
Can you tell us if the problem is still present also with the current last release 1.6.2? Thanks!

kdeyko:
Hi @robertobandini, thank you for your reply.

Sadly, both mentioned issues do persist in 1.6.2.

PR #2338, opened 2026-07-25
Improve macOS terminal text navigation

joyheroes:
Map macOS natural text editing shortcuts in the embedded terminal so zsh receives the same word navigation and line kill sequences as native terminals.

Constraint: Issue #1051 reports macOS ZSH behavior in the embedded terminal, while existing terminal key handling is renderer-local.

Rejected: Changing PTY or shell startup behavior | the mismatch is caused by renderer key translation before data reaches the shell.

Confidence: medium

Scope-risk: narrow

Directive: Keep these mappings macOS-only and avoid intercepting shifted or otherwise modified variants unless a separate issue proves they are expected.

Tested: pnpm exec vitest run -c vitest.config.ts --project @freelensapp/core --pool forks --reporter verbose packages/core/src/renderer/components/dock/terminal/terminal.test.ts; pnpm biome check packages/core/src/renderer/components/dock/terminal/terminal.ts packages/core/src/renderer/components/dock/terminal/terminal-key-mapping.ts packages/core/src/renderer/components/dock/terminal/terminal.test.ts; pnpm type:check

Not-tested: Manual Freelens terminal smoke test in the Electron app.


dex4er:
@joyheroes Thanks for the patch. I verified on my Mac, and it really works with zsh, so I'm happy that someone fixed it.

joyheroes:
Thanks for checking it on your Mac! Glad it works well with zsh. Happy I could help fix this :)

Issue #1237, opened 2025-09-25
restarted containers with yellow border can be confused with not-ready one


giuliocalzolari:
### What would you like to be added?

After https://github.com/freelensapp/freelens/pull/1004 

with dark theme is not so straight forward to understand if the pod is healthy or restarted/not ready

<img width="371" height="215" alt="Image" src="https://github.com/user-attachments/assets/30698232-9cfd-4c92-bd4b-90ef2d0a65ef" />

is it possible to change the color of the border using grey or white instead of orange??

### Why is this needed?

This ui improvement could help to immediately visualize the problematic pods

robertobandini:
I remember that there is another issue about this topic, I mean about that this orange border maybe is not the best solution, but in this moment I can't find the issue.

Since several people have pointed this out, I'd say it's still feedback worth considering.

Anyway, there is another issue/idea (really from the beginning of Freelens I can say) about to choose your own colors.
So this could be another color that anyone can edit in your favorite theme:
https://github.com/freelensapp/freelens/issues/550#issuecomment-3325730461

What do you think?

FYI @mariomamo

Cenness:
> can't find the issue

It was discussion. https://github.com/freelensapp/freelens/discussions/1103

dex4er:
This is an indicator of the problems with the pod. Usually, restarts of the pods are real issues that need fixing. The normal remediation should be fixing the manifest, then restarting the deployment/statefulset, and it'll make the green indicator come back.

I'm listening for the comments, but I don't understand yet what is confusing about this indicator regarding restarted containers? Generally = yellow or orange colors present some kind of warning, and this is exactly the warning about an unhealthy situation.

tplueer:
I was also confused by this, as it was very hard to tell whether the container was currently running or still initializing.

I think it’s just a matter of improving how restarts are displayed—for example, by increasing the spacing around the border.

Something like that would already make it clearer for me.

```
.StatusBrick.restarted {
    background-color: var(--colorOk);
    outline: 1px solid #ff8200;
    outline-offset: 3px;
```

<img width="1084" height="174" alt="Image" src="https://github.com/user-attachments/assets/87853e2a-8141-4384-836d-28de83273a9b" />

robertobandini:
> > can't find the issue
> 
> It was discussion. [#1103](https://github.com/freelensapp/freelens/discussions/1103)

Thanks @Cenness

mnorrsken:
[...]

Issue #186, opened 2025-02-26
After the filter box is selected, it turns white, making it difficult to see clearly in a light - colored theme.


arcanique:

robertobandini:
Hi @arcanique and thank you about your contribute!
Can you provide a description and a screenshot so everybody can understand more clearly?
Thank you very much!

jtechtips:
similar to me, when use Light theme, when select any text, the selected text turns white, difficult to see clearly.

<img width="778" alt="Image" src="https://github.com/user-attachments/assets/be0eee1a-cb18-4312-908b-6473c4458c3e" />
select "version" text

<img width="757" alt="Image" src="https://github.com/user-attachments/assets/29db6d1b-2867-4f71-930f-f35f9f4964dd" />

robertobandini:
> similar to me, when use Light theme, when select any text, the selected text turns white, difficult to see clearly.

[...]

PR #2349, opened 2026-07-26
Add custom accent color preference

jearthliu:
This adds a custom accent color setting to the appearance preferences, letting users override the primary/blue accent colors. Its a small change — a color picker + hex input + reset button — but it addresses the most common pain point people have with the current themes (issues #550, #1237, #186, #1280).

The override is stored in user preferences and applied as CSS custom properties after the active theme loads, so it works with any theme. Happy to adjust the approach if the maintainers have a different preference for how this should work.


robertobandini:
Hi @jearthliu,

As explained in #1280 (see my comment there), the issue is still assigned to
gambhirsharma, so no further PRs are accepted, whether from humans or bots.
You can find our bounty policy here: #2293

For this reason I'm closing this PR without review.

In any case, everyone is welcome here! There are many ways to contribute to
Freelens and its extensions: open issues, documentation, the extensions
ecosystem, and of course other bounties when available. Take a look at
CONTRIBUTING.md and feel free to join us.

Thank you for your interest in Freelens!

PR #2351, opened 2026-07-26
Verify every kubectl download against a pinned checksum

### LINKED ISSUE #2348: Pin and verify checksums for kubectl versions downloaded at runtime
## Problem

Issue #2342 closed the supply-chain gap for the three **bundled** binaries: their digests
are pinned in `freelens/binaries.lock.json` and every publisher signature is verified when a
version is bumped.

The application still downloads kubectl on its own at runtime, and that path has no
verification at all. `Kubectl.downloadKubectl()` in
`packages/core/src/main/kubectl/kubectl.ts` writes whatever the URL returns straight to
disk. The only check that follows is `checkBinary()`, which runs `kubectl version --client`
and compares `gitVersion`. A substituted binary reporting the expected version passes that
check without difficulty.

The download URL is built from the mirror selected in preferences, so today the trust
boundary ends at TLS to `dl.k8s.io` -- or at whatever host a user typed into the custom
mirror field.

Goal: pin a verified sha256 for every kubectl the application may download, verify it at
download time, and refuse anything unpinned.

## What upstream offers

Measured against `dl.k8s.io` on 2026-07-26, for the latest patch of each minor in the
version map:

| Range | `SHA256SUMS` | `kubectl.sig` + `kubectl.cert` |
| --- | --- | --- |
| 1.22 - 1.36 (15 minors) | yes | **yes** |
| 1.13 - 1.21 (9 minors) | yes | no |
| 1.0 - 1.12 (13 minors) | no | no |

`cosign verify-blob` was run against real downloads at both ends of the range and in the
middle -- v1.22.17, v1.23.17, v1.24.17 and v1.36.3 all return `Verified OK` using the
constants already present in `packages/ensure-binaries/src/verify.ts`:

```
--certificate-identity    krel-staging@k8s-releng-prod.iam.gserviceaccount.com
--certificate-oidc-issuer https://accounts.google.com
```

In every case the computed digest also matched the published `.sha256`. The signing
identity is stable across the range and does not need versioning, so
`verifyKubectlSignature` is reusable unchanged.

**Trap worth recording.** Each release also publishes `provenance.json` (an in-toto SLSA
v0.2 statement listing every artifact with its digest) and a single `SHA256SUMS` covering
all binaries. Both look like a cheap way to pin every variant from one small fetch. Neither
is signed -- there is no `.sig` or `.cert` for either file, and `provenance.json` is a bare
statement rather than a DSSE envelope. Trusting them is trusting TLS, which is the thing
being removed. Digests must be computed from bytes we actually downloaded. (`v1.33.13` also
publishes no `provenance.json` at all while still publishing signatures, so depending on it
would be fragile regardless.)

## Scope: kubectl 1.22 and newer

1.22 is a hard floor rather than a preference: 1.21.14 publishes no signature, so any pin
below 1.22 would rest on the vendor checksum alone. It is also the lowest such line, which
makes it the choice that maximises cluster compatibility without conceding anything on
verification -- every pin is signed either way, and every pin below the currently supported
minors is frozen either way.

Ongoing cost of reaching down to 1.22 is zero: 1.22.17 and 1.23.17 are the final patches of
EOL branches, so the nightly job will never touch them again.

Note this is deliberately *not* aligned with the `Kubernetes >= 1.24` declared by
[freelens-example-extension](https://github.com/freelensapp/freelens-example-extension#requirements),
which governs CRD API versions in an example extension -- a different question from kubectl
download policy. An extension may require more than the application does; the two numbers
are not in conflict and neither should be "fixed" to match the other.

**Variant coverage is not uniform.** All six platform/architecture combinations exist for
1.23 and later, but **v1.22.17 has no `windows/arm64`** (404). That is 89 pins, not 90, and
the generator must treat a missing variant as data rather than as an error.

Only x64 and arm64 are built (see `freelens/electron-builder.yml`), so there are at most six
variants per version: darwin, linux and windows against amd64 and arm64. The `386` branch in
`kubectl-downloading-normalized-arch.injectable.ts` is unreachable in shipped builds and is
not pinned.

## Plan

Two pull requests. The first is inert -- it adds pins that nothing reads yet. The second
turns enforcement on. This keeps the large backfill diff separate from the behavioural
change.

### PR 1 -- generate and commit the pins (#2350, merged)

**`packages/kubectl-versions/build/compute-versions.mts`** -- ~~cut the map at 1.22~~
**deferred to PR 2.** Cutting the map is not inert on its own: with `kubectlVersionMap` no
longer naming a minor below 1.22, `Kubectl.ts` falls through to its `coerce()` fallback for
those clusters -- trying the cluster's own exact patch version instead of the table's latest
known patch -- which is a real behavior change with nothing yet to consume the pins that
would make it safe. `update-kubectl-checksums.ts` got its own `MIN_SUPPORTED_MINOR = 22`
instead, so the generator stays correctly scoped regardless of what `versions.json` lists,
and #2350 changes nothing observable at runtime. The map cut belongs in PR 2, alongside the
refusal-plus-bundled-fallback that makes the gap deliberate.

**`packages/ensure-binaries/src/update-kubectl-checksums.ts`** (new) -- the generator.
Reuse, do not duplicate:

- `describeArtifact({tool: "kubectl", version, platform, arch})` from `artifacts.ts`, which
  already builds the canonical `dl.k8s.io` URL and the `.sha256` sidecar URL
- `artifactKey()` from the same file, which yields `${platform}/${arch}`, matching
  `binaries.lock.json`
- `fetchChecksum` and `downloadToFile` from `download.js`
- `verifyKubectlSignature` via `verifyArtifact` in `verify.js`
- `mapWithConcurrency(items, 4, ...)` from `update-lock.ts`

The per-artifact flow in `pinArtifact` (`update-lock.ts:78`) is exactly what is needed --
`mkdtemp`, download, compare against the published checksum, verify the signature, `rm` in
`finally`. Extract it into a shared module rather than copying it.

Read from `dl.k8s.io` only, never from a mirror. This is a correctness requirement rather
than a preference: pinning bytes fetched from a mirror would let a compromised mirror bless
its own digest, whereas pinning from the canonical source is what holds every mirror to what
upstream actually published. The mirror preference applies to the runtime download only.

Skip artifacts already present in the file. That makes the generator incremental and makes
an existing pin immutable by construction.

Treat a 404 on the artifact itself as "this variant was never published" -- record nothing
and carry on -- but **log every skipped variant**, so an unpublished gap cannot quietly read
as full coverage. A 404 on the `.sha256` or on the signature of an artifact that *does*
exist must remain a hard error.

**`packages/kubectl-versions/build/checksums.json`** (new, committed) -- keyed by version,
then by `${platform}/${arch}`, mirroring the lock shape. Sorted and stable, so an unchanged
set regenerates byte-identically -- the CI diff check depends on that, exactly as it does
for `binaries.lock.json`.

**`packages/kubectl-versions/index.ts`** -- exports `kubectlChecksums` alongside
`kubectlVersions`. `versions.json` stays untouched, per the deferral above.

**`.github/workflows/kubectl-versions.yaml`** -- generator step added after
`compute-versions`, timeout raised 25 -> 45 minutes.

**`.github/workflows/kubectl-checksums-check.yaml`** (new) -- modelled on
`binaries-lock-check.yaml`. Verifies only entries added relative to the base commit, and
asserts no existing digest changed.

**Also fixed in #2350:** `trunk-fmt-pre-commit` was permanently blocking any commit that
touched a generated `build/*.json` file (reproduced independently of this issue, on
`build/tsconfig.json`), because Trunk's biome linter definition kept handing biome paths
that `biome.jsonc` itself excludes, and biome 2.5.4 treats that as a hard error rather than
a no-op. Fixed with a `lint.ignore` entry plus `--no-errors-on-unmatched` on both biome
commands in `.trunk/trunk.yaml`.

### PR 2 -- enforce at download time, and the rest of this issue (#2351, merged)

**`packages/core/src/main/kubectl/kubectl-checksums.injectable.ts`** (new) -- wraps
`kubectlChecksums` and exposes a lookup by `(version, platform, arch)`, wired through
`create-kubectl.injectable.ts` and registered in `register-injectables.ts`.

**`packages/kubectl-versions/build/compute-versions.mts`** -- cut the map at 1.22, now that
enforcement makes the resulting gap deliberate rather than a silent behavior change.

**`packages/core/src/main/kubectl/kubectl.ts`**:

- `ensureKubectl()` refuses when no pin exists for the resolved version/platform/arch,
  naming the `kubectlBinariesPath` preference instead of the previous generic "Failed to
  ensure kubectl".
- `downloadKubectl()` hashes the in-memory buffer against the pin -- no checksum fetched at
  runtime -- and writes through a temporary file (`${this.path}.download-<random>`, never
  given an `.exe` suffix) that only earns a `chmod` and an atomic `rename` once the digest
  matches. The previously leaked file handle is closed, and the rename retries on
  `EPERM`/`EBUSY`.
- `kubectl.test.ts` (new): 8 unit tests over the match, mismatch, refusal and cleanup paths.

**`preferences-helpers.ts`** -- `chinaPackageMirror` removed. `mirror.azure.cn/kubernetes/kubectl/`
stopped syncing around August 2024; no replacement exists. No migration needed --
`downloadMirror`'s descriptor already coerces an unknown stored value back to the default.

**`README.md`** -- a short `## Requirements` section: Kubernetes 1.22 or later, older
clusters fall back to the bundled kubectl.

**`AGENTS.md`** -- documents the pin file and how to regenerate it.

**Also fixed alongside #2351:** 14 workflows shared
`echo "pnpm_cache_dir=$(pnpm store path)" | tee -a ${GITHUB_ENV}`, which silently swallows a
failing `pnpm store path` -- only `echo`'s own exit status decides whether the step fails, so
a transient Corepack fetch failure produced an empty `pnpm_cache_dir` that surfaced two steps
later as an unrelated `actions/cache` "Input required and not supplied: path" error. Fixed by
assigning to a shell variable first, so `set -e` aborts where the real failure is.

## Cost

kubectl ranges from roughly 45 MB (v1.23.17) to 60 MB (v1.36.3) per artifact. Eighty-nine
pins is about 4.5 GB as a one-off backfill.

That figure is **cumulative network transfer, not disk usage**. The generator downloads,
hashes and deletes one artifact at a time, so peak disk is roughly `concurrency x 60 MB` --
around 240 MB at a concurrency of 4.

Steady state is far cheaper. EOL minors are frozen, so only supported ones move; a typical
patch release day touches three minors, or 18 downloads. The nightly `kubectl-versions.yaml`
job's timeout was raised to 45 minutes in #2350 to cover this.

## Verification

Done, in CI and locally: `pnpm type:check`, `biome check`, `trunk check` clean; core unit
suite passing, including the 8 new `kubectl.test.ts` cases covering digest match, mismatch
(content rejected, nothing left behind, existing binary untouched), unpinned-version refusal
and unpinned-arch refusal.

**Not run against a live cluster or a hostile mirror** -- the two runtime scenarios below
are exercised by `kubectl.test.ts`'s mocked `downloadBinary` (a mismatching response is
functionally the same input as a hostile mirror), but neither was performed end-to-end
against a real cluster:

- a cluster below 1.22, confirming the fallback and the log message naming
  `kubectlBinariesPath`
- opening Preferences with `downloadMirror: "china"` still in an existing user store and
  confirming the select shows Default (the coercion itself predates this issue and is
  untested either way)

Neither blocks closing this issue -- both are low-risk, already covered in spirit by unit
tests -- but worth a manual pass before relying on them.

## Tasks

PR 1 (#2350, merged):

- [x] ~~Cut `compute-versions.mts` at 1.22~~ -- deferred to PR 2, see above
- [x] Extract the `pinArtifact` flow from `update-lock.ts` into a shared module
- [x] Add `update-kubectl-checksums.ts`, reading from `dl.k8s.io` only, skipping existing
      entries and logging unpublished variants
- [x] Backfill `packages/kubectl-versions/build/checksums.json` for 1.22 - 1.36 (89 entries)
- [x] Export `kubectlChecksums` from `packages/kubectl-versions/index.ts`
- [x] Add the generator step to `.github/workflows/kubectl-versions.yaml` and recheck the
      timeout
- [x] Add `.github/workflows/kubectl-checksums-check.yaml` for added and modified pins

PR 2 (#2351, merged):

- [x] Cut `compute-versions.mts` at 1.22
- [x] Add `kubectl-checksums.injectable.ts` and wire it through `create-kubectl.injectable.ts`
- [x] Refuse unpinned version/platform/arch combinations in `ensureKubectl()`, with a message
      pointing at `kubectlBinariesPath`
- [x] Rewrite `downloadKubectl()` with the temp-file, verify, rename sequence, keeping the
      temporary name free of an `.exe` suffix
- [x] Close the file handle and add the Windows rename retry
- [x] Remove `chinaPackageMirror` from `preferences-helpers.ts`
- [x] Add a `## Requirements` section to `README.md` stating Kubernetes 1.22 or later
- [x] Document the pin file and its regeneration in `AGENTS.md`

## Out of scope

Attestation for kubectl below 1.22, which upstream does not publish. Helm and
freelens-k8s-proxy are not downloaded at runtime and are already covered by #2342.

dex4er:
PR 2 of the plan in #2348, covering the rest of the issue. PR 1 (#2350) added the pins; this turns enforcement on.

## What changed

- `packages/kubectl-versions/build/compute-versions.mts` -- the map starts at 1.22, matching the generator's floor; `versions.json` regenerated to 15 minors.
- `packages/core/src/main/kubectl/kubectl-checksums.injectable.ts` (new) -- lookup by (version, platform, arch), wired through `create-kubectl.injectable.ts`.
- `packages/core/src/main/kubectl/kubectl.ts` -- `ensureKubectl()` refuses an unpinned version/platform/arch and names the `kubectlBinariesPath` preference; `downloadKubectl()` hashes the in-memory buffer against the pin and writes through a temporary file that only a matching digest earns a `chmod` and a rename from. The leaked file handle is closed, and the rename retries on `EPERM`/`EBUSY`.
- `packages/core/src/main/kubectl/kubectl.test.ts` (new) -- 8 tests over the match, mismatch and refusal paths.
- `preferences-helpers.ts` -- `chinaPackageMirror` removed; the mirror stopped syncing around August 2024.
- `README.md` -- a `## Requirements` section: Kubernetes 1.22 or later.
- `AGENTS.md` -- the pin file and how to regenerate it.

## Note for review

`ensureDir(this.dirname)` stays above the refusal: `binDir()` writes the shell session's init scripts into that directory whether or not a download happens, and those scripts are what put the bundled binaries directory on the terminal's PATH.

## Verification

- `pnpm type:check`, `biome check`, `trunk check --upstream main` clean.
- Core unit suite: 2367 passed / 27 skipped.
- The runtime checks in the issue's Verification section need a live cluster and a hostile mirror, and were not run here.

Fixes #2348

Generated with [Claude Code](https://claude.ai/code) | Model: `claude-opus-5`


dex4er:
@claude check why https://github.com/freelensapp/freelens/actions/runs/30208766756/job/89811338795?pr=2351 failed

PR #2373, opened 2026-07-27
Use Chromium fetch in the renderer and undici in main

### LINKED ISSUE #2361: Migrate the fetch layer from node-fetch to undici
## Goal

Replace `node-fetch` 3 with `undici` across the app and the public `Common.K8sApi`
surface, removing `hpagent` along the way.

Split out of the dependency audit in #2360 so the transport rewrite does not block the
catalog cleanup.

## Current state

`node-fetch` 3.3.2 (published 2023-07-25) is a runtime dependency of `packages/core` and
`freelens`, and a type dependency of the published `@freelensapp/extensions`
(`extensions/common-api/k8s-api.ts:51,143` exposes `NodeFetchRequestInit`;
`JsonApi` / `KubeJsonApi` expose `RequestInit` / `Response`).

## The transport differs per process

The first attempt at this migration (#2371, closed) used undici in **both** processes and
crashed in the renderer. Two things were measured on the running dev app and on
undici 8.9.0; both contradict the original plan in this issue.

### undici does not run in the renderer

The renderer is Chromium with `nodeIntegration`: `require` and `process` are Node's, but
`setTimeout` and `performance` are the page's. Measured in the running cluster frame:

```text
timerType: "number"    hasUnref: undefined
hasMarkResourceTiming: undefined
hasRequire: "function" nodeVersion: "24.18.0"
```

undici needs both — its fast-timer loop calls `.unref()` on the result of the global
`setTimeout`, and `fetch` calls `performance.markResourceTiming` after every request. So
the cluster frame spews, once per second and once per request:

```text
TypeError: fastNowTimeout?.unref is not a function       (undici timers, every tick)
TypeError: markResourceTiming is not a function          (undici fetch, every response)
```

This is not a bundling artifact — loading undici through Node's `require` instead of the
Vite prebundle changes nothing, because a module required in the renderer shares the
page's globals. **undici is a main-process-only dependency.**

### The `Host` header is not a blocker in the renderer — it is the frame's own origin

`renderer/k8s/api-kube.injectable.ts` and `common/k8s-api/api-base.injectable.ts` connect
to `https://127.0.0.1:<lensProxyPort>` and then hand-set `Host: window.location.host`.
That is the header the browser would send anyway if the request went to the frame's own
origin: `setup-hostnames.injectable.ts:22` maps `renderer.freelens.app` and
`*.renderer.freelens.app` to 127.0.0.1 in Chromium's host resolver,
`setup-lens-proxy-certificate.injectable.ts:40` puts both in the certificate's SANs, and
`create-electron-window.injectable.ts:114` makes the session trust it.

Verified from the cluster frame in the running app, with plain Chromium `fetch` and no
header tricks:

```text
origin: https://<clusterId>.renderer.freelens.app:58301
fetch(`${location.origin}/api-kube/version`) -> 200
{"major":"1","minor":"35", ..., "gitVersion":"v1.35.6-gke.1127000", ...}
```

So the renderer needs no `Host` header, no CA dispatcher (the session already trusts the
lens-proxy certificate) and no Node fetch implementation at all.

### undici's `fetch` drops `Host` too

The claim in the original version of this issue — that undici allows setting `Host` where
Chromium does not — is wrong. Measured on undici 8.9.0 against a local HTTP server:

| client | `Host: a.b.c` reaches the server |
| --- | --- |
| `node-fetch@3` | yes |
| `undici.fetch` (plain object, `Headers`, upper or lower case) | **no** — rewritten to the URL authority |
| `undici.request` (core dispatch API) | yes |

`fetch`'s `Request` constructor applies the Fetch spec's "request" header guard, which
strips `host` in undici exactly as it does in Chromium. Only the core dispatch layer
honours it.

**Main still needs the header**, because Node has no equivalent of Chromium's
host-resolver rules: `main/k8s/api-base-host-header.injectable.ts` and
`common/k8s-api/create-kube-api-for-cluster.injectable.ts:67` route to a cluster with
`Host: <uid>.renderer.freelens.app:<port>` over a socket to 127.0.0.1. So main keeps a
dispatcher interceptor that re-adds the header below the `fetch` guard —
`host-header-dispatcher.ts` from #2371 is the part of it worth keeping.

### `hpagent` is deleted, not vendored

`hpagent` 1.2.0 (2022-10-29) has exactly one usage —
`main/fetch/https-agent.injectable.ts:8`, main process only, `HttpsProxyAgent` for the
`httpsProxy` user preference. undici ships `ProxyAgent`, so hpagent falls out for free.
(The `overrides: http-proxy: npm:http-proxy-node16` entry is unrelated — that is the proxy
*server*.)

## Tasks

- [ ] Point the renderer at its own origin: `renderer/k8s/api-base-server-address.injectable.ts`
      returns the frame origin instead of `https://127.0.0.1:<port>`, and the `Host` header
      drops out of `renderer/k8s/api-kube.injectable.ts` and the renderer path of
      `common/k8s-api/api-base.injectable.ts`.
- [ ] Give the fetch injectable a per-process implementation: Chromium's global `fetch` in
      the renderer, undici's in main. Keep the injectable ids and the global overrides
      tests rely on. undici must not be imported from a module the renderer bundles.
- [ ] Keep the `Host` smuggling for main only (`host-header-dispatcher.ts`), covered by a
      test that drives a real undici request against a local server.
- [ ] Convert `main/fetch/https-agent.injectable.ts` to an undici `Agent` / `ProxyAgent`
      dispatcher and drop `hpagent` from `packages/core` and `freelens`.
- [ ] Update `JsonApi` / `KubeJsonApi` (`utility-features/json-api`, `utility-features/kube-api`)
      to types both implementations satisfy, and drop the per-request keep-alive agents
      (undici pools per dispatcher; Chromium pools on its own).
- [ ] Rename or re-type the exported `NodeFetchRequestInit` in
      `extensions/common-api/k8s-api.ts` and drop `node-fetch` from `catalogs.extensions`
      and the default catalog.
- [ ] Rewrite the watch-stream line splitter in `kube-api.ts:914` for a WHATWG
      `ReadableStream` body — this is the `byline` removal from #2360. Both transports
      deliver one, so this is shared.
- [ ] Migrate the build-time `node-fetch` users (`ensure-binaries`, the kubectl
      version-map generator) so the catalogs can lose the dependency.
- [ ] Verify live: renderer reaches clusters through lens-proxy, and an HTTPS proxy
      configured in preferences still works.

## Risk notes

- The cluster routing is the load-bearing behaviour and unit tests with a mocked fetch will
  not catch a regression. Live smoke test required, in both processes.
- Response body semantics differ from node-fetch (Node `Readable` versus WHATWG
  `ReadableStream`), which affects the watch stream and any code that consumes
  `response.body` directly.
- The renderer and main now return different `Response` implementations (DOM versus
  undici). They are structurally compatible for everything `JsonApi` does, but the shared
  types have to be written so that neither process needs a cast.
- This is extension-API-breaking (`Common.K8sApi` request/response types) and should be
  announced together with the other breaks in #2301.

Split out of #2360. First attempt: #2371 (closed, wrong premise).

dex4er:
Closes #2361. Supersedes #2371, which used undici in both processes and crashed the renderer. Rebased onto #2375, which removed `forRemoteCluster`.

Replaces `node-fetch` 3 with a per-process transport, and removes `hpagent` and `byline` along the way.

## The two processes cannot share a fetch client

**undici does not run in the renderer.** Its fast-timer loop calls `.unref()` on the result of the global `setTimeout`, and its `fetch` calls `performance.markResourceTiming` after every request. The renderer's globals are Chromium's, measured in the running cluster frame:

```text
timerType: "number"    hasUnref: undefined
hasMarkResourceTiming: undefined
hasRequire: "function" nodeVersion: "24.18.0"
```

A module required in the renderer shares the page's globals, so loading undici through `require` rather than the bundler does not help — which is what #2371 ran into, once per second and once per response:

```text
TypeError: fastNowTimeout?.unref is not a function
TypeError: markResourceTiming is not a function
```

**The renderer needs no Node client anyway.** `api-kube` and `api-base` used to connect to `https://127.0.0.1:<lensProxyPort>` and hand-set `Host: window.location.host` — the header the browser sends by itself once the request goes to the frame's own origin. `*.renderer.freelens.app` is mapped to 127.0.0.1 by Chromium's host-resolver rules (`setup-hostnames.injectable.ts`), is in the lens-proxy certificate's SANs (`setup-lens-proxy-certificate.injectable.ts`), and the window's session already trusts it (`create-electron-window.injectable.ts`). So the renderer now requests its own origin with Chromium's `fetch`, and both the `Host` header and the CA-carrying agent fall away.

**Main keeps a Node client**, because Node resolves no such hostname: it connects to 127.0.0.1 and names the target cluster itself. undici's `fetch` drops a `Host` header just like Chromium's — the Fetch spec's request header guard strips it in both, and only the core dispatch layer honours it:

| client | `Host: a.b.c` reaches the server |
| --- | --- |
| `node-fetch@3` | yes |
| `undici.fetch` (record, entry list or `Headers`; any case) | **no** — rewritten to the URL authority |
| `undici.request` | yes |

`main/fetch/host-header-dispatcher.ts` lifts the header out of the request and re-adds it from a `Dispatcher.compose()` interceptor, below the guard. Composed dispatchers are cached per (dispatcher, host), since `JsonApi` sends the same `Host` on every request. Its test drives a **real** undici request against a local `http.Server`, so this cannot regress behind a mocked fetch.

## What changed

- **`@freelensapp/json-api` owns the shared contract** (`src/fetch.ts`): the DOM request/response types both clients satisfy, plus the `dispatcher` undici extension main needs and Chromium ignores. `JsonApi`, `KubeJsonApi` and `KubeApi` take those types. The module-level keep-alive `http.Agent`/`https.Agent` are gone — undici pools per dispatcher, Chromium pools on its own.
- **`common/fetch/fetch.injectable.ts`** is the one fetch the app injects: a façade over a per-process implementation token, so no module the renderer bundles imports undici. `main/fetch/undici-fetch.injectable.ts` and `renderer/fetch/browser-fetch.injectable.ts` implement it; the latter replaces the unused `renderer/fetch/fetch.injectable.ts` left over from #924. The lens-proxy dispatcher is a token for the same reason, resolving to `undefined` in the renderer.
- **`K8sApi.forCluster` and `KubeJsonApi.forCluster` address the target cluster by URL in the renderer.** They reach a cluster other than the one whose frame the caller runs in, and named it in a `Host` header — which Chromium drops, so the request would have landed on the frame's own origin and silently answered for the wrong cluster. `clusterApiAddressInjectionToken` now answers per process: the lens-proxy address plus the `Host` header in main, the cluster's own origin (`https://<clusterId>.<lens-proxy authority>`) in the renderer. Covered by a test, because the failure is silent.
- **`main/fetch/https-agent.injectable.ts`** returns an undici `Agent`, or a `ProxyAgent` when the `httpsProxy` preference is set; `hpagent` is gone. Because undici pools per dispatcher, it is now built once and rebuilt only when the preferences it derives from change, closing the superseded one — the old code allocated a fresh `https.Agent` per request.
- **The watch-stream line splitter** reads the WHATWG `ReadableStream` body with a reader and reassembles the newline framing itself, replacing `byline`: a chunk can hold several events and an event can straddle two chunks. End-of-stream and stream errors drive the same retry the `"end"`/`"close"`/`"error"` events used to.
- **`common/utils/get-latest-version.injectable.ts` moves to `main/versions/`.** It is used only from main (the renderer goes through a channel), but importing `main/fetch/download-json` from `common/` pulled the whole of undici into the renderer bundle.
- **`window.location.host` carries the port**, so `windowLocationInjectable` no longer returns a separate `port` — its only consumer was the address the renderer now derives from the origin. The test fake said `host: "localhost"`, `port: "12345"`, which no real browser would.
- The **build scripts** (`ensure-binaries`, the kubectl version-map generator) move to undici in the same commit rather than a separate one, because `node-fetch` can only leave the catalogs — and the lockfile — once its last consumer does.

## Breaking changes for extensions

`Common.K8sApi` request/response types are the DOM's, extended with `dispatcher`; `NodeFetchRequestInit` becomes `FetchRequestInit`.

That is the whole list. An earlier revision of this PR also changed `CreateKubeApiForRemoteClusterConfig.agent` into a `.dispatcher`, and had to note that the remote cluster's TLS options would silently stop working in the renderer. #2375 removed `forRemoteCluster` outright, so neither applies any more — and the last commit here drops the `tlsDispatcherInjectionToken` that existed only to serve it.

Should be announced together with the other breaks in #2301.

## Verification

`pnpm type:check`, `pnpm test:unit` (3519 passed, 44 skipped), `pnpm biome check`, `pnpm trunk check`, `pnpm build` — all green. **undici is absent from the renderer bundle** and external in main, asserted against the built sourcemaps rather than by inspection.

### Live smoke test

Done on a GKE cluster (v1.35.6-gke.1127000) through `pnpm dev`, driving the real app:

- **Cluster views load through the renderer's Chromium fetch.** Cluster connects; the overview renders node warnings and events; Pods lists 140 items, Deployments 77, all reaching the cluster over `https://<clusterId>.renderer.freelens.app:<port>/api-kube/...` with no `Host` header set by hand. Pod logs stream too.
- **Watch works through the new splitter.** `?watch=1` requests stay open, and the Events view grew 188 → 200 items over 45 seconds with no reload and **zero console errors**. (Before, #2371 produced hundreds of errors per minute in the same frame.)
- **Host routing**, checked against the running proxy from outside the app: with `Host: <clusterId>.renderer.freelens.app:<port>` the request reaches the cluster; without it, the app shell answers.
- **The HTTPS proxy preference works through undici's `ProxyAgent`.** With a logging CONNECT proxy on 127.0.0.1:8899 set in Preferences → Proxy, main's npm-registry version check tunnelled through it — `CONNECT registry.npmjs.org:443` — and completed without error. The preference was cleared again afterwards.

| Model: `claude-opus-5[1m]`


dex4er:
Pushed 61609c49, which fixes a cross-cluster misroute the first commit introduced.

`K8sApi.forCluster` and `KubeJsonApi.forCluster` reach a cluster **other** than the one whose frame the caller runs in, and they did it by naming the target in a `Host` header while connecting to the lens-proxy address. Once the renderer moved to Chromium's fetch that header is dropped, so every such call landed on the frame's own origin — silently answering for the current cluster, or for the app shell when called from the root frame. Requests kept succeeding, which is what makes it worth a test rather than a comment.

Both factories now ask `clusterApiAddressInjectionToken` where the cluster lives:

- **main** — the lens-proxy address plus the `Host` header it has always used, carried through by the dispatcher interceptor;
- **renderer** — the cluster's own origin, `https://<clusterId>.<lens-proxy authority>`, which Chromium can address directly (host-resolver rules map every `*.renderer.freelens.app` to 127.0.0.1 and the certificate covers them all). The authority comes from the current frame's host with any cluster-id label removed, so it holds in a cluster frame and in the root frame alike.

`pnpm type:check`, `pnpm test:unit` (3523 passed), `pnpm biome check` and `pnpm trunk check` are green.

Worth noting for the reviewer how this differs from the `forRemoteCluster` limitation already listed under breaking changes: `forCluster` targets a cluster **in the catalog**, so lens-proxy has a `freelens-k8s-proxy` running for it and the renderer only needs to address it correctly — fixed here. `forRemoteCluster` takes an ad-hoc server URL that is in no catalog, has no proxy instance behind it, and never goes through lens-proxy at all, which is why its TLS options cannot be honoured from the renderer.

dex4er:
@claude rebase after #2375 is merged

dex4er:
Rebased onto #2375 and pushed 8ada15e3, which finishes what the ordering note in #2374 was for.

The rebase resolved the conflict on `create-kube-api-for-remote-cluster.injectable.ts` correctly — the file stays deleted — but it left `tlsDispatcherInjectionToken` behind as dead code. That injectable was the token's only injection site, so what remained was the token, its main implementation, its renderer no-op and two generated registrations, serving nobody. Nothing flags that on its own: it compiles, the registrations resolve, the tests pass. Removed by hand, with `pnpm build:di` rerun for the two `register-injectables.ts` files.

The breaking-changes section of the description has shrunk accordingly: the only extension-visible change left is the `Common.K8sApi` request/response types (`NodeFetchRequestInit` → `FetchRequestInit`). The `.agent` → `.dispatcher` change and the caveat about TLS options going quiet in the renderer both died with `forRemoteCluster`.

`pnpm type:check`, `pnpm test:unit` (3519 passed, 44 skipped), `pnpm biome check` and `pnpm trunk check` green on the rebased branch. The live smoke test in the description was run before the rebase; nothing in #2375 or 8ada15e3 touches the paths it exercised, so I have not repeated it.

dex4er:
@claude fix knip check

dex4er:
@claude check https://github.com/freelensapp/freelens/actions/runs/30345697058/job/90231258598?pr=2373 integration tests result and analyze stacktrace

dex4er:
@claude apply fix

PR #2380, opened 2026-07-31
docs: add community Fedora COPR installation

kappa8219:
## Proposal

Can I move to this org https://github.com/kappa8219/freelens-copr and repackages the official Freelens RPM


kappa8219:
@dex4er what do you think?

Issue #2360, opened 2026-07-26
Prune the @freelensapp/extensions dependency surface before the v2 spec freezes it


dex4er:
## Goal

Remove dead and ill-fitting third-party packages from the published
`@freelensapp/extensions` type surface, and from the app tree behind it, before #2304
turns the &#34;Third-party bundled libraries&#34; contract into normative specification. Every
entry in `catalogs.extensions` is a promise to extension authors that is hard to walk back
once the spec is frozen.

This also closes the one item Phase 3 (#2278) explicitly deferred: `react-table` 7.

### There are no pending majors

Comparing all 138 catalog pins against `dist-tags.latest` yields four major gaps, none of
them React 19 debt:

| Package | Pin | npm latest | Assessment |
| --- | --- | --- | --- |
| `@types/node` | 24.12.4 | 26.1.1 | deliberate — Electron 42 is Node 24 |
| `electron` | 42.7.1 | 43.2.0 | routine Renovate bump |
| `chalk` | 5.6.2 | 6.0.0 | main process only |
| `electron-vite` | 6.0.0-beta.1 | 5.0.0 | we are *ahead* of stable |

The whole React / MobX / testing family is current. The problem is not outdated versions,
it is packages that are current **and abandoned** — which a major-version diff cannot
surface by construction.

### `react-table` 7

| | Value |
| --- | --- |
| `react-table` 7.8.0 | published **2022-05-16**; repo moved to TanStack, v7 frozen |
| peer range | `react: ^16.8.3 \|\| ^17.0.0-0 \|\| ^18.0.0` — **no 19** |
| lockfile | `react-table@7.8.0(react@19.2.8)` — resolved despite the unmet peer |
| renderer bundle cost | `react-table.production.min.js` = **57,969 B** |
| public API type cost | `@types/react-table` 7.7.20, **920 lines** of DefinitelyTyped `index.d.ts` (frozen) |
| our code | `table/react-table.tsx` 95 lines + `list/list.tsx` 44 lines + `react-table.scss` 46 lines |
| consuming screens | **one** — `renderer/components/extensions/installed-extensions.tsx` |

Surface actually used: `useTable` + `useFlexLayout` + `useSortBy`; columns with `Header` /
`accessor` / `width` / `sortType` / `disableSortBy`; prop getters that reduce to inline
`style` (flex) and `onClick` (sort toggle). No pagination, grouping, filtering (`List`
filters itself) or virtualization. The flex layout is already in SCSS
[...]

PR #2390, opened 2026-07-31
Replace stoppable with the Node server close built-ins

dex4er:
Part of #2360 (third item of task list A2).

**Description of changes:**

- Replace `stoppable(server, 500)` in the lens proxy with `server.close()` + `closeIdleConnections()` + `closeAllConnections()` on a timer.
- Drop `stoppable` and `@types/stoppable` from `packages/core`, `freelens`, the catalog and the lockfile.
- Add tests for `LensProxy#close`, which had none.

`stoppable` was last released in 2018-11. Its `stop()` means "stop accepting, close the idle connections, force whatever is left after the grace period" — Node has done all three since 18.2, and we run 24. The `stoppable.WithStop` intersection leaves the `proxyServer` field along with it.

**Why this one is load bearing rather than cosmetic**

The proxy carries connections that never go idle — watch and follow requests, and the sockets of upgraded shell sessions — and `closeLensProxyOnQuitInjectable` awaits `close()` during the quit sequence. A close that waits for those to end on their own does not return, so the forced destroy after the grace period is the part that has to keep working.

I read `stoppable`'s implementation rather than trusting the summary in #2360. Two differences worth recording, neither of which changes the outcome here:

- Node ≥19's `server.close()` already closes idle connections, so `closeIdleConnections()` is belt-and-braces. It is kept because it makes the intent legible next to the grace timer.
- `stoppable`'s forced path calls `socket.end()` and then `socket.destroy()` a tick later, where `closeAllConnections()` destroys immediately. At quit that distinction does not matter.

**Tests**

Driven against a real listening server, using the existing `node:https` → `node:http` mock in this file:

- resolves with nothing connected
- resolves without waiting out the grace period when a connection is idle
- destroys a connection whose request never answers, once the grace period is up
- second call is a no-op

The third test needed a `routerInjectable` override that never responds. My first attempt just wrote partial headers and never finished the request, on the assumption that made the connection busy — it does not. Node still counts that connection as idle, `closeIdleConnections()` reaped it, and the test passed for the wrong reason without touching the forced path at all.

**Validation**

- `pnpm type:check` — clean
- `biome check` on both touched files — clean
- `trunk check` on the changed config files — clean
- `src/main/lens-proxy/` — 18 passed

Not covered: the real TLS server and real upgraded websocket sockets, since the unit test swaps https for http. The quit path is worth exercising once in the running app.


dex4er:
@claude check why unit test failed and make a fix

PR #2414, opened 2026-08-11
fix: prevent multiple file/folder dialogs from opening simultaneously

### LINKED ISSUE #1850: Multiple File/Folder Dialogs Open When Clicking + Button Repeatedly
### Describe the bug

In the Clusters Catalog (and possibly other areas with a + button), clicking the + button opens a file/folder selection dialog as expected. However, if the user clicks the + button multiple times quickly, multiple dialogs open simultaneously. This can confuse users and is not a professional UI behavior.

### Reproduction steps

- Open Freelens (1.9.1-0) on Windows.
- Navigate to Catalog → Clusters.
- Click the green + button. The file/folder dialog opens.
- Click the + button again before closing the dialog.
- Observe that a second dialog opens on top of the first.

### Expected behavior

- Only one dialog should be open per + button action.
- While a dialog is open, additional clicks should be ignored or queued until the current dialog closes.

### Screenshots

_No response_

### Freelens version

1.9.1-0

_No response_

### Installation method

_No response_

_No response_

### Logs

_No response_

### Kubeconfig

_No response_

### Additional context

_No response_

waterWang:
## Description

When clicking the + button in the Clusters Catalog (or any path-picking button) multiple times quickly, multiple file/folder selection dialogs open simultaneously. This creates a confusing UX.

## Changes

1. **openPathPickingDialog injectable**: Added a guard flag that prevents opening a new dialog while one is already open. Subsequent calls are silently ignored until the current dialog closes.
2. **PathPicker component**: Added a disabled state while the dialog is open, providing visual feedback and preventing accidental repeated clicks.

## Fixes

Fixes #1850

### DIFF
--- a/packages/core/src/features/path-picking-dialog/renderer/pick-paths.injectable.ts
+++ b/packages/core/src/features/path-picking-dialog/renderer/pick-paths.injectable.ts
@@ -16,15 +16,26 @@ const openPathPickingDialogInjectable = getInjectable({
   id: "open-path-picking-dialog",
   instantiate: (di): OpenPathPickingDialog => {
     const requestFromChannel = di.inject(requestFromChannelInjectionToken);
+    let isDialogOpen = false;
 
     return async (options) => {
-      const { onPick, onCancel, ...dialogOptions } = options;
-      const response = await requestFromChannel(openPathPickingDialogChannel, dialogOptions);
+      if (isDialogOpen) {
+        return;
+      }
 
-      if (response.canceled) {
-        await onCancel?.();
-      } else {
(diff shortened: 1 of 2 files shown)


robertobandini:
Hi @waterWang! Thanks for this fix, the approach is the right one and I could reproduce the issue on macOS too.

To move it forward without waiting, I opened #2455 which keeps your commit as is and adds on top of it a small change (the guard is released as soon as the native dialog closes, so a long running onPick does not block the other file dialogs of the app) plus a unit test. You will be credited as co-author when it is merged.

I am closing this one in favour of #2455.

PR #2415, opened 2026-08-11
fix: keep cluster metrics history on clusters with node churn

### LINKED ISSUE #2409: Cluster metrics show almost no history on autoscaled clusters
### Describe the bug

The Cluster overview CPU/Memory charts build their PromQL node selector from the
nodes alive at query time, but take the time window from the metrics range
picker. On clusters with high node churn (Karpenter, Cluster Autoscaler, spot)
nearly every node from earlier in the window has been replaced, so its samples
are filtered out even though Prometheus still holds them.

Selecting a 5-day range gives a chart that is empty for ~4 days and only rises
near the right edge — which reads as lost history or an idle cluster. The step
ramp at the edge is just the current nodes coming into existence one by one.

`cluster-metrics.injectable.ts` snapshots the live node list, then
`request-cluster-metrics-by-node-names.injectable.ts` joins it into the
`node=~"..."` selector in the providers. The node filter is "now" while
start/end can be days back.

### Reproduction steps

1. Use a cluster that recycles nodes frequently (ours replaces them several
   times a day) with Prometheus retention longer than a node's lifetime (5d).
2. Cluster page > Worker > CPU, metrics range 5 days.
3. Data appears only for roughly the age of the oldest surviving node.

Same 6-day window queried directly, filtered to the currently-live nodes as
Freelens does, vs. unfiltered:

```
sum(rate(node_cpu_seconds_total{node=~"<live nodes>", mode=~"user|system"}[5m]))
  -> 53 of 124 hourly points, meaningful only for the last ~18h
sum(rate(node_cpu_seconds_total{mode=~"user|system"}[5m]))
  -> 124 of 124, full window
```

475 distinct nodes appear in that window; 13 are alive now, the oldest 19h old.
Retention is healthy (`prometheus_tsdb_size_retentions_total` 0, no gaps in
`count(up)`), so this is purely the selector.

### Expected behavior

A 5-day chart should show 5 days, including terminated nodes — the panel
aggregates with `sum()` over the cluster, so their past contribution is part of
the answer.

Expressing the master/worker split in PromQL (joining node label metrics) rather
than as a snapshot of names would make the filter evaluate per timestamp, which
is what the time axis already implies. Dropping the filter for the cluster-wide
aggregate is simpler but loses the split.

### Screenshots

metrics exists for 5 days

Image

### Freelens version

1.10.3

macOS 26.5.2

### Installation method

DMG

1.35 (EKS)

### Logs

_No response_

### Kubeconfig

_No response_

### Additional context

In-cluster prometheus-community/prometheus chart (Prometheus v3.12.0), queried
directly — no Grafana in the path.

The time-range picker from #1962 is what surfaced this; before long ranges were
selectable the live-node snapshot was close enough to hide it. Probably affects
upstream Lens too, unverified.

waterWang:
### Description

Fixes #2409

The cluster overview CPU/Memory charts build their PromQL node selector from the nodes alive at query time, but take the time window from the metrics range picker. On clusters with high node churn (Karpenter, Cluster Autoscaler, spot) nearly every node from earlier in the window has been replaced, so its samples are filtered out even though Prometheus still holds them.

### Fix

For cluster-level metrics, only filter by node names when a specific list is provided. The cluster overview now passes an empty node list, which tells the PromQL providers to omit the node label selector — aggregating all nodes over the time range.

### Changes

- **cluster-metrics.injectable.ts**: pass empty node list instead of current nodes
- **request-cluster-metrics-by-node-names.injectable.ts**: only set opts.nodes when nodeNames is non-empty
- **provider.ts**: add nodeFilter / nodeFilterAdditive helper functions
- **All 4 PromQL providers** (lens, helm, operator, stacklight): use conditional node filter

### How to Test

1. Connect to a cluster with autoscaler (Karpenter, Cluster Autoscaler) or node churn
2. Open Cluster overview → CPU/Memory charts
3. Select a time range spanning back before the current nodes were alive
4. Verify the chart shows historical data instead of being empty/ramp

### DIFF
--- a/packages/core/src/common/k8s-api/endpoints/metrics.api/request-cluster-metrics-by-node-names.injectable.ts
+++ b/packages/core/src/common/k8s-api/endpoints/metrics.api/request-cluster-metrics-by-node-names.injectable.ts
@@ -58,10 +58,16 @@ const requestClusterMetricsByNodeNamesInjectable = getInjectable({
     const requestMetrics = di.inject(requestMetricsInjectable);
 
     return (nodeNames, params = {}) => {
-      const opts = {
+      const opts: Record<string, string> = {
         category: "cluster",
-        nodes: nodeNames.join("|"),
       };
+
+      // Only filter by node names when a specific list is provided.
+      // When empty, cluster-wide metrics aggregate across all nodes,
+      // preserving historical data from replaced nodes (see #2409).
+      if (nodeNames.length > 0) {
+        opts.nodes = nodeNames.join("|");
+      }
       const { metrics = defaultClusterMetricKeys, ...requestParams } = params;
       const query = metrics.reduce(
         (acc, metricName) => {
--- a/packages/core/src/renderer/components/cluster/cluster-metrics.injectable.ts
+++ b/packages/core/src/renderer/components/cluster/cluster-metrics.injectable.ts
@@ -9,7 +9,6 @@ import { now } from "mobx-utils";
 import requestClusterMetricsByNodeNamesInjectable from "../../../common/k8s-api/endpoints/metrics.api/request-cluster-metrics-by-node-names.injectable";
 import { asyncComputed } from "../../../common/utils/async-computed";
 import selectedMetricsTimeRangeInjectable from "./overview/selected-metrics-time-range.injectable";
-import selectedNodeRoleForMetricsInjectable from "./overview/selected-node-role-for-metrics.injectable";
 
 import type { ClusterMetricData } from "../../../common/k8s-api/endpoints/metrics.api/request-cluster-metrics-by-node-names.injectable";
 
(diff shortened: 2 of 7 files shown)


robertobandini:
Hi @waterWang! Thanks for taking this one, the idea of omitting the node selector when no node list is given is the right starting point.

Two things needed a follow-up: the Prometheus Operator provider needs its kube_pod_info join when a node list is given (the Nodes page requests the same queries for a single node, and node-exporter series have no node label in that setup), and the cluster overview still has to honour the Master/Worker choice when both roles exist. I opened #2456 which keeps your commit as is and adds those fixes plus tests on top of it, so you will be credited as co-author when it is merged.

Closing this one in favour of #2456.

PR #2417, opened 2026-08-15
Fix the broken star history chart in the README

FaintFlower:
The star history chart embedded in the README is currently broken and no longer renders, so visitors lose a quick overview of the project's growth. This change moves the chart to a working service so it displays again. No other README content is affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ Anyone is welcome to request to be added to the official list by opening a dedic
 Anyone is welcome to collaborate to advance the Freelens project. Read
 [CONTRIBUTING.md](CONTRIBUTING.md) to see how you can help.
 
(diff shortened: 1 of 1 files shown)


robertobandini:
Thanks for taking the time to open this.

We are going to keep the README pointed at star-history.com and wait for GitHub's fix rather than switch providers. The chart is broken because GitHub restricted the stargazers API to repository admins and collaborators on 2026-06-30, and the star-history maintainers report that GitHub is actively working on restoring the previous behavior.

We also prefer not to embed an image from a third-party service that is not affiliated with star-history (see star-history/star-history#552). In the meantime we will hide the chart from the README and restore it once the API is available again.

Closing this one.

PR #2426, opened 2026-08-22
Register alternative keyboard shortcuts for zooming in

### LINKED ISSUE #1848: Zoom In Keyboard Shortcut (Ctrl + +) Not Working
### Describe the bug

In version 1.9.1-0, the View → Zoom functionality behaves inconsistently:

Zoom Out / Ctrl + - works correctly with both mouse menu and keyboard shortcut.
Zoom In / Ctrl + + only works when clicking the menu item with the mouse.
Pressing the keyboard shortcut (Ctrl + +) does not scale the UI.

### Reproduction steps

1. Open Freelens (1.9.1-0) on Windows.
2. Press Ctrl + +.
3. Observe that the UI does not zoom in.
4. Click View → Zoom In with the mouse. The UI zooms in correctly.

### Expected behavior

1. The UI should scale larger when using Zoom In from the keyboard shortcut.
2. Keyboard and mouse zoom controls should behave consistently.

### Screenshots

_No response_

### Freelens version

1.9.1-0

Windows 11

### Installation method

_No response_

_No response_

### Logs

_No response_

### Kubeconfig

_No response_

### Additional context

_No response_

robertobandini:
Zoom In could not be triggered from the keyboard on layouts where `+` is not the shifted `=`.

## Root cause

Electron's `zoomIn` role defaults to the `CommandOrControl+Plus` accelerator. In Electron 42.7.1 (`shell/common/keyboard_util.cc`) the `plus` token maps to `{VKEY_OEM_PLUS, '+'}` with a *shifted* character, so `accelerator_util.cc` adds `EF_SHIFT_DOWN`: the accelerator is literally **Ctrl + Shift + VKEY_OEM_PLUS**. That only matches keyboards where `+` is the shifted `=` (US/UK). On Italian, German, Nordic and similar layouts the key Windows reports as `VK_OEM_PLUS` is the dedicated `+` key, pressed *without* Shift, so the combination never matches; the numeric keypad `+` (`VKEY_ADD`) never matches on any layout. Zoom Out is unaffected because `CommandOrControl+-` needs no Shift. Clicking the menu item bypasses accelerator matching, which is why the mouse path works.

## Fix

An Electron `MenuItem` holds a single accelerator, so the alternatives are registered as **hidden** menu items: `GenerateAcceleratorTable` walks the menu model without looking at visibility (it only honours `registerAccelerator`, which is `true` for the zoom roles), so a `visible: false` item keeps its shortcut without adding a duplicate entry to the View menu.

- `OsActionMenuItem` gains an optional `visible` flag (`MayBeHiddenFromMenu`), passed through to the Electron template
- Zoom In gains `CommandOrControl+=` (the unshifted `VKEY_OEM_PLUS`: `+` on IT/DE keyboards, `=` on US ones, also the common Chrome-style shortcut) and `CommandOrControl+numadd`
- Zoom Out gains `CommandOrControl+numsub`
- regression test `zoom-keyboard-shortcuts.test.ts` (one visible item per direction, alias accelerators present), snapshot and DI registration updated

Applies to v2 (`main`); the same three source files are byte-identical on `v1`, so a backport is a clean cherry-pick (the new test needs porting from vitest to jest).

## Verification

Run locally on macOS (x64) against this branch, replicating the PR workflows:

- unit tests: 352 files / 3591 tests passed (full suite), including the new test and the win32/darwin/linux menu snapshots
- `type:check`, `build:di` (no drift), `knip:check`, `trunk check --all`: clean
- packaged app (`electron-builder --macos dir`) + integration tests: passed (KinD-dependent tests skipped as in CI)

**Not yet verified: the actual key presses on Windows.** The unit test only proves the Electron menu template carries the accelerators; the matching happens in native code and depends on the keyboard layout. Someone with Windows and an Italian/German layout should check Ctrl + `+` (main key), Ctrl + keypad `+`, Ctrl + keypad `-`, and on a US layout that Ctrl+Shift+= still works and Ctrl+= now works too. If nobody gets to it, I will do it myself ASAP.

Fixes #1848

| Model: `claude-opus-5` (analysis and fix), `claude-fable-5` (review and verification)


robertobandini:
I built this branch on Windows (portable Node 24.18.0, no elevated access needed) and tested the packaged app personally with the **Italian QWERTY** keyboard layout: I confirm the fix works — Zoom In is now triggered correctly from the keyboard.

This covers the "not yet verified on Windows" part of the PR description.

As per our usual practice, I'm requesting a review from @dex4er before merging.

PR #2427, opened 2026-08-24
fix(#1110): show metrics for init containers

### LINKED ISSUE #1110: Show metrics for init containers

When clicking on a single pod one can see metrics like CPU, Memory and Filesystem on Containers. It would be nice if Init Containers also showed this.

Container example:

Image

How Init Container currently looks:

Image

### Why is this needed?

Some init containers like istio-proxy are vital to the pods operations, and it is very useful to be able to inspect its CPU and Memory usage in FreeLens.

prakhardewangan2005-hash:
Fixes #1110

Init container metrics were never rendered because the metrics block in pod-details-container.tsx was gated on containersType === "containers". Init containers carry type: "initContainers", so the condition excluded them.

podContainerMetricsInjectable already keys on ${pod.getId()}-${container.name} and queries with the pod, container, namespace selector, both of which resolve fine for init containers — so no changes were needed beyond widening the condition.

Ephemeral containers are intentionally left out of this change.

Note: I don't currently have a working local build environment, so this hasn't been verified by running the app. Happy to hold this until it can be validated, or wait to be assigned first per #2293 if that's preferred.

### DIFF
--- a/packages/core/src/renderer/components/workloads-pods/pod-details-container.tsx
+++ b/packages/core/src/renderer/components/workloads-pods/pod-details-container.tsx
@@ -116,7 +116,7 @@ class NonInjectedPodDetailsContainer extends React.Component<PodDetailsContainer
           <StatusBrick className={containerStatusClassName(container, status)} />
           {name}
         </div>
-        {isMetricVisible && containersType === "containers" && (
+        {isMetricVisible && (containersType === "containers" || containersType === "initContainers") && (
           <>
             {containerMetrics.map((ContainerMetrics) => (
               <ContainerMetrics.Component key={ContainerMetrics.id} container={container} pod={pod} />


prakhardewangan2005-hash:
Update after reviewing the metrics layer more closely: this change is incomplete as-is.

The Prometheus providers (lens, helm, operator, stacklight) all query kube_pod_container_resource_requests / _limits. kube-state-metrics publishes init-container requests and limits under separate kube_pod_init_container_resource_requests / _limits series, so with this change an init container's chart would render usage but leave the request/limit lines empty — which is arguably more misleading than showing nothing.

Two options I can see:

Make the query init-container-aware — pass the container type through to the provider layer and pick the metric name accordingly. Touches all four providers.
Render init-container metrics for usage only, and hide the request/limit series for that case.

I'd lean toward (1) since it's the correct fix, but it's a wider change than the one-liner here. Happy to take either direction — @robertobandini which would you prefer?

Issue #2366, opened 2026-07-27
Enforce the extension API surface with an API report


dex4er:
## Goal

Turn the v2 extension API surface from something described in prose into something CI
checks: a git-tracked API report, plus a mechanical gate against the class of defect
found in #2365.

Depends on #2363 — API Extractor has to be running first. Split out of it so this value
does not hide inside a build-fix issue; the two have different urgency and different
readers.

## Why prose is not enough here

#2304 is meant to produce the *normative* specification of the extension API, and one of
its open contract items is "namespace enumeration" — what each of `Common` / `Main` /
`Renderer` actually provides. Written by hand, that list is stale the first time someone
adds an export.

The audit in #2360 and the gap in #2365 both came out of reading the surface manually.
That does not scale, and more importantly it does not repeat: nothing stops the next
export from reintroducing the same problem.

There is a structural reason this matters more for v2 than it would for a normal library.
Workspace packages are private and inlined into the bundled `.d.ts` (39 entries; the
runtime half is a four-line shim over `globalThis.FreelensExtensionApi`). With that model,
**a symbol the API namespace does not re-export is unreachable by any means** — there is no
package left for an author to install. Completeness of the re-export is therefore a
correctness property of the packaging model, not a convenience.

### 1. `ae-forgotten-export` — the mechanical check for #2365

Reported when an exported API references a declaration that is not itself exported from
the entry point. That is exactly the defect #2365 documents by hand:

| Exported by the API | Its signature needs | Nameable by an extension? |
| --- | --- | --- |
| `KubeApi` | `KubeApiOptions & ExternalKubeApiOptions` | `ExternalKubeApiOptions` yes, **`KubeApiOptions` no** |
| `PodsApi`, `NodesApi`, … | `DerivedKubeApiOptions` | **no** |
| `KubeObjectStore` | `KubeObjectStoreOptions` | **no** |

By default this message goes to the report file rather than the console, so intentional
exceptions stay documented and reviewable instead of being silenced.

### 2. The `.api.md` report — a diffable contract

A git-tracked file containing the public API surface. From the API Extractor docs: *"The
report file should be tracked by Git, so that changes to an API signature will appear as
diffs when a pull request (PR) is created."*

This is the artifact #2304 actually wants. Instead of a hand-maintained namespace
enumeration that drifts, the enumeration *is* the checked-in file, and any change to it
shows up in review. The rushstack repository gates this with `.github/CODEOWNERS` so API
[...]

Issue #2363, opened 2026-07-26
@freelensapp/extensions build:dist is broken: TypeScript 7 removed the JS compiler API


dex4er:
## Summary

`pnpm --filter @freelensapp/extensions build:dist` fails on `main`. Since `build:dist` is
what `prepack` runs, **the published `@freelensapp/extensions` package cannot currently be
built or packed.**

Found while verifying the dependency audit in #2360, which needs the bundled
`extension-api.d.ts` as its source of truth for what the package really imposes on
extension authors.

> **Revised diagnosis.** This issue originally blamed `rollup-plugin-dts` for not
> supporting TypeScript 7, and listed two failures. Both were understated: the root cause
> is that **TypeScript 7 removed the classic JavaScript compiler API**, and *three* of the
> pipeline's steps are affected, not two. The tool choice below changed accordingly.

## Root cause: TypeScript 7 has no JS compiler API

`typescript@7.0.2` no longer exports `ts.*`. Its `exports["."]` points at
`./lib/version.cjs`, and `lib/` contains only `tsc.js`, `getExePath.js` and `version.cjs`.
Everything else lives under an explicitly-unstable surface (`./unstable/sync`,
`./unstable/async`, `./unstable/ast`, …).

Verified at runtime:

```
$ node -e "import('typescript').then(m=>{console.log(Object.keys(m.default));
    console.log('transpileModule?', typeof m.default.transpileModule);
    console.log('sys?', typeof m.default.sys)})"
[ 'version', 'versionMajorMinor' ]
transpileModule? undefined
sys?             undefined
```

Any tool that reaches for the compiler API is therefore dead on TypeScript 7, regardless
of its own maintenance status. This is not a `rollup-plugin-dts` defect.

## Three affected steps, not one

| Step | Tool | Status |
| --- | --- | --- |
| `build:dist:declarations` | `tsc` CLI | runs, but reports 5 TS2883 errors |
| `build:dist:bundle` | `rollup-plugin-dts` → `ts.sys` | **dead** |
| `build:dist:shim` | `scripts/build-shim.mjs` → `ts.transpileModule` | **dead** |

### Failure 1 — declaration emit reports TS2883

`tsc -p tsconfig.dts.generated.json` exits 1 with five errors of the same shape:

```
../core/src/main/catalog-sync-to-renderer/broadcaster.injectable.ts(14,7): error TS2883:
  The inferred type of 'catalogSyncBroadcasterInjectable' cannot be named without a
[...]

PR #2446, opened 2026-09-09
Fix the @freelensapp/extensions build:dist pipeline on TypeScript 7

dex4er:
`pnpm --filter @freelensapp/extensions build:dist` has been broken since #2202 bumped TypeScript to v7: the release ships no JavaScript compiler API, and two of the three steps reached for it. Since `build:dist` is what `prepack` runs, the published package could not be built or packed.

Implements option 4 from the [analysis on #2363](https://github.com/freelensapp/freelens/issues/2363#issuecomment-5607184903). `typescript` stays `catalog:` = 7.0.2 everywhere; no override, no `peerDependencyRules`, no `dlx`, no new workspace package.

### Commits

1. **Emit the runtime shim without the compiler API** - `scripts/build-shim.mjs` ran `src/runtime-shim.ts` through `ts.transpileModule`. The only TypeScript syntax in the shim is two `!` non-null assertions, so a text substitution replaces the compiler. The shim stays a `.ts` file under `pnpm type:check` and no build step is added. Two guards keep it honest: the script fails unless it finds exactly two assertions, and `node --check` parses the emitted file as ESM.
2. **`rollup-plugin-dts` 6.4.1 -> 6.5.1** - 6.5.1 (2026-08-04) is the plugin's TypeScript 7 support release. It declares `typescript: ^4.5 || ^5 || ^6 || ^7` and probes for the compiler API at runtime. This also resolves the previously unmet peer.
3. **`@typescript/typescript6` 6.0.2** - the vendor bridge Microsoft published for this gap, and the plugin's optional peer. Default catalog plus `devDependencies` of `packages/extensions`, deliberately *not* `catalogs.extensions`: it is a build-time bridge and must not appear in what extension authors resolve. Retired once the plugin supports the TypeScript 7.1 API (microsoft/TypeScript#63703, stable 2026-11-24).
4. **`noEmitOnError` in `tsconfig.dts.json`** - a failing declaration emit no longer leaves a half-written `dist-types/` for the rollup to bundle.
5. **A CI gate** - see the note below.

### Verification

- `pnpm --filter @freelensapp/extensions build:dist` succeeds from a clean tree, all three steps.
- `dist/extension-api.d.ts` is **byte-identical** to the TypeScript 7 reference: sha256 `1711e596c7ed88b62e02e537332a88962a582b3eb6785bc7dbbef2feaadfc9dc`, 13038 lines, the same 20 external import statements.
- `dist/extension-api.js` differs from the old `transpileModule` output by **three blank lines only** - the compiler collapsed the blank lines between the header comments and the statements; a copy keeps them. The code is character for character the same. Flagged rather than papered over, since the definition of done asked for byte identity.
- `pnpm type:check` green, `pnpm peers check` reports no issues, `biome check` clean, `trunk check` clean on the changed files, and `knip` output is byte-identical to the baseline in both its development and production pass (the bridge is not flagged, so no ignore entry was needed).

### A maintainer must move the workflow file by hand

The agent's token lacks the `workflows` permission, so the CI gate lands as a complete file at `github-workflow-fix/extensions-build-dist.yaml`. **It must be moved to `.github/workflows/extensions-build-dist.yaml` manually, unchanged, for the gate to take effect.** It was verified with actionlint 1.7.12 from a real `.github/workflows/` path. Until then this PR fixes the pipeline but does not prevent it from rotting again.

### Out of scope

The undeclared `es-toolkit` external and the double `child_process` / `node:child_process` spelling now visible in the produced declaration belong to #2360; the API report to #2366.

Refs #2363

| Model: `claude-opus-5[1m]`


dex4er:
@claude Please fix knip test regression and remove from comments references to our issues and too verbose rationales. Also, comments in .json file don't seem to be right.

PR #2462, opened 2026-09-17
fix: make "Download > All logs" work again in production builds

robertobandini:
"Download > All logs" in the logs dock tab does nothing in production builds of the current main, for any pod: no request, no notification, no file. "Visible logs" works.

Cause: `download-logs-dropdown.tsx` wrapped the call in `act` from `react-dom/test-utils`, a line inherited from OpenLens. In React 19 that module delegates to `React.act`, which is not exported by the production bundle, so the click throws `TypeError: React.act is not a function` before `downloadAllLogs` runs. Checked against the workspace React 19.2.8:

```
NODE_ENV=production node -e 'console.log(typeof require("react").act)'
undefined
```

Fix: call `downloadAllLogs()` directly. `act` is a test-only helper and the `waiting` state around the call is unchanged.

Found while testing #2418 end to end, where the same click failed on a plain single-pod tab too. The existing suite in `features/pod-logs/download-logs.test.tsx` has the "download all logs" cases skipped (TODO on the menu item test id), which is why this went unnoticed; the skip is left untouched here and deserves its own look.

Verification: type check, biome, the pod-logs and dock/logs unit tests, and a manual check with the unpackaged build on a kind cluster: "All logs" now downloads the full log of the pod (109 lines), "Visible logs" unchanged.

Generated with [Claude Code](https://claude.com/claude-code)


AleksaMCode:
@robertobandini, would it be possible to add tests for this so regressions can be caught before merging a PR (I see there are `download-logs.test.tsx` and `download-logs-dropdown.tsx`)? I’m guessing I’m not the only one relying on logs, and it’s pretty annoying when you can’t access them due to bugs in Freelens.

PR #2463, opened 2026-09-17
Cancel stale pod log requests and prevent polling backlogs

Sungyoun-Kim:
**Description of changes:**

Pod log polling can accumulate pending requests while the selected Pod is missing from the renderer's store. Each 10-second tick starts another asynchronous `loadMore` call, and each call waits independently for Pod/tab data. When that data becomes available again, all pending calls request the same logs at once. A separate race allows an initial load to restart polling after the user has left its tab.

- Allow at most one pending load or poll per tab, including the wait for Pod/tab data.
- Cancel both the MobX wait and the HTTP request when a tab stops or a new load supersedes the current request.
- Ignore late results and errors from canceled requests so they cannot restart polling or overwrite newer logs.
- Forward an optional `AbortSignal` through `CallForLogs` and `PodApi.getLogs`, preserving existing callers.
- Add regression tests for request accumulation, slow responses, tab cleanup, reloads, independent tabs, and loading older lines; also verify signal forwarding to fetch.

**Reproduction:**

The new `LogStore` tests reproduce the behavior without a Kubernetes cluster:

1. Load a tab successfully so its polling interval starts.
2. Make the selected Pod unavailable in the observable store.
3. Advance fake timers by one hour, then restore the Pod.

Before this change, 360 pending polls issue log requests when the Pod returns. With this change, only one pending poll resumes. Another test stops a tab before its initial response completes and verifies that the late response does not restart polling. All 10 new lifecycle regression cases fail against the original implementation and pass with this change.

**Validation:**

- Log viewer, pod-log feature, and Kubernetes API tests: 110 passed, 15 skipped (13 test files passed, 1 skipped).
- `pnpm type:check`: passed.
- `pnpm build:fixture-extension`: passed.
- `pnpm biome check` on all five changed files: passed.
- Trunk 1.25.0 check on all five changed files in a Linux/WSL checkout of the same base and patch: passed.
- Full Vitest suite on Windows: 3,562 passed, 35 failed, 44 skipped (349 test files passed, 8 failed, 1 skipped). Re-running all eight failing files with the three production source files restored to the base implementation reproduced exactly the same 35 failing test names. These are existing Windows-specific path separator snapshots, shell PATH expectations, and a POSIX executable-mode assertion; no unrelated snapshots or platform code are changed here.

Validated locally on Windows with Node.js 24.19.0. No live-cluster or Electron end-to-end validation was performed.


robertobandini:
Hi @Sungyoun-Kim, thank you for this one, and welcome!

I approved the workflows (all green) and went through it locally: type check, biome, the logs and kube-api test suites, plus a run of the unpackaged build on a kind cluster with the namespace filter toggled while a logs tab was open. Nothing changed in the normal path, one poll every 10 seconds as before, no errors. The backlog itself I could only cover through your regression tests, since a Pod that disappears from the store does not come back with the same id in a real cluster, so thanks for making them reproduce it without a cluster.

Two small notes, neither blocking: `when` with an AbortSignal is a first in this code base, fine with MobX 6.15, and `node:assert` in renderer code matches what we already do elsewhere.

One heads-up: #2418 (combined logs for workloads) rewrites the same functions of the LogStore and the two branches conflict. My plan is to get this one in first, since it is small and self-contained, and ask the author of #2418 to rebase on it and reuse your AbortSignal for the per-pod requests. I have asked @dex4er for the review.

Sungyoun-Kim:
Thanks @robertobandini for the thorough review and for testing this on a kind cluster!

For some context on what prompted this fix, we investigated a production incident where Freelens 1.9.0 sent 1,419 identical pod-log requests within approximately 2.58 seconds, with 1,557 requests observed in total. We verified these counts using Kubernetes audit logs, deduplicated by audit ID.

Shortly afterward, API server logs recorded:

```text
rpc error: code = ResourceExhausted desc = etcdserver: throttle: too many requests
```

Karpenter and EBS CSI event-write requests also received HTTP 429 responses with this message. EBS CSI sidecars exited after failing to renew their leader-election leases, and the Karpenter controller restarted.

The timing suggests that the burst contributed to the control-plane incident. Although we haven't reproduced the full incident end to end, the repeated requests for the same Pod with identical query parameters are consistent with the backlog behavior covered by the regression tests.

